PR #61, opened 2015-05-30
add functions, tests for complex numbers

joostvanzwieten:


gertjanvanzwieten:
Thanks for picking this up!
Some comments. Calling real or imag on a function this way always wraps the function object in a Pointwise object regardless of dtype, not returning self or zero if the function is real valued. Pointwise is a bit of an shallow construct anyhow, used for fast wrapping of mathematical functions like sin or ln that don't need to define any special methods for function tree manipulation (operation swaps, simplifications).
I would say that real and imag justify proper ArrayFuncs, that can define a _localgradient (localgradient(real(f)) -> real(localgradient(f))) and other manipulations like _get, _inflate, etc. The abs method can then be defined as sqrt( real(arg)**2 + imag(arg)**2 ), which (with a small modification in Power to avoid recursion) reduces to arg * sign(arg) for real valued arguments, and yields the correct gradient in all cases.
Could you add that? I think these are relatively minor modifications on top of the work you already did.

joostvanzwieten:
Actually, I think we should disallow computing gradients of functions at complex values. The complex derivatives of `abs`, `real` and `imag` are not defined. Since it is not possible to determine whether a function argument is real or complex when building a function tree, maybe we should add a `localcomplexgradient` and dedicated complex gradients to all functions. Nevertheless, the suggested implementation for `abs` makes sense, so I will send a v2.

gertjanvanzwieten:
Why is it not possible to determine real or complex on build? The dtype is a property of the ArrayFunc so I'd say we have that information. We just need to update all operations such as Add and Multiply to correctly set the dtype of the result based on their arguments. That way, multiplying a real valued function with a complex number results in a complex Multiply object. Calling real on that pushes the operation (through Real._multiply) down to its arguments, where it vanishes on the function and turns the number to real. Likewise for localgradient. And having that accurate dtype logic also sets int * int to int, and int * float to float; that's much needed functionality.

joostvanzwieten:
See #64.

PR #76, opened 2015-09-24
log video v4

joostvanzwieten:
See the docstring of `PyPlotVideo` for an explanation. I use the numpy format for the docstrings, but with an indentation of two spaces instead of four.


joostvanzwieten:
See #77.

PR #77, opened 2015-09-24
log video v5

joostvanzwieten:
See the docstring of `PyPlotVideo` for an explanation. I use the numpy format for the docstrings, but with an indentation of two spaces instead of four.

Compared with #76 child processes are now properly closed when the user kills the simulation (13d873c) and time-dependent examples can generate videos when running with `--plt=video` (625e1e6).

### DIFF
--- a/examples/advection
+++ b/examples/advection
@@ -3,9 +3,11 @@
 from nutils import *
 
 
-def makeplots( domain, geom, u ):
+def makeplots( domain, geom, u, plt ):
   xp, up = domain.elem_eval( [ geom, u ], ischeme='bezier9', separate=True )
-  with plot.PyPlot( 'solution', ndigits=4 ) as plt:
+  if plt is True:
+    plt = plot.PyPlot( 'solution', ndigits=4 )
+  with plt:
     plt.mesh( xp, up )
     if domain.ndims == 2:
       plt.gca().set_aspect( 'equal' )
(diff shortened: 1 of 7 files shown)


gertjanvanzwieten:
Looks good! The only question I have is unrelated to video: why did you add 'frames is None' to util.run? The main point of waiting for subprocesses to end is to not have their output overrunning the traceback explorer, which exactly applies to when frames is not None. I'm sure the answer is in 'deadlock' in the comment but I didn't get that.

joostvanzwieten:
Ah, now I get understand why you were waiting for subprocesses to die. Well, keeping the stack alive might keep subprocesses alive, resulting in a deadlock when waiting for them to die. What about killing subprocesses iff the traceback explorer is started?

joostvanzwieten:
Something else. Shall I automatically 'focus' a figure when entering a `with`-block? This eliminates the `plt._pyplot.figure(plt._fig.number)` lines in 625e1e6.

gertjanvanzwieten:
That would be much better, except I couldn't think of a way to do that without maintaining a global list of child pids. Until now, that is: I can probably kill them from Fork.__exit__ directly. Let me look into that.

gertjanvanzwieten:
Regarding focusing: absolutely. That should be the entire purpose of the with-block as far as I'm concerned.

joostvanzwieten:
We could also [redirect stdout] of subprocesses when starting the traceback explorer.

[redirect stdout]: http://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/

gertjanvanzwieten:
Interesting. Never considered that. We might want to do both: killing child processes from within Fork, and muting output prior to starting the explorer just to be safe. This is out of the video scope though, is it an idea to remove the util.run modification and pull in the rest, then fix the child issue?

joostvanzwieten:
See #78.

PR #83, opened 2015-11-21
Misc

joostvanzwieten:
Four unrelated commits.

### DIFF
--- a/nutils/cache.py
+++ b/nutils/cache.py
@@ -57,7 +57,11 @@ def argdict( f ):
     return value
   return f_wrapped
 
-class HashableArray( object ):
+class HashableBase( object ):
+
+  pass
+
+class HashableArray( HashableBase ):
   # FRAGILE: assumes contents are not going to be changed
   def __init__( self, array ):
     self.array = array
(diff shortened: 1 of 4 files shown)


gertjanvanzwieten:
Looks good! One question, in the new blockadd function you removed argument sorting, in fact you introduce an OrderedDict to hold on the original order of entry. The reason for sorting is to have a unique final Blockadd object (for de-duplication and easy detection of cancellations etc.) Was there a reason to remove it?

gertjanvanzwieten:
Another thing, in _test_pickle_dump_load, it is unclear to me why we need to involve subprocess to test pickling/unpickling. Why not just dumps/loads?

joostvanzwieten:
As `BlockAdd` automatically drifts to the top, there is probably only one `BlockAdd` in the function tree, or one per concatenate item, unless trapped in some non-linear function. And I was too lazy to implement sorting for more than two arguments. But it's a one-liner so I can easily add this if you want.

Now we are talking about `_sorted`, I dislike the runtime randomness. I'll open an issue to discuss this as soon as I have a reasonable alternative.

My first solution for pickling actually worked when unpickling in the same process. Pickled objects were taken from `cache.Immutable`'s cache instead of being recreated, and that's where another bug was hidden.

gertjanvanzwieten:
I fully agree about runtime randomness so let's discuss ways to fix that. Also, maybe we can add a cache flush for testing, I'll make a note of that. We'll go with this solution for now.

PR #98, opened 2016-03-01
Deriv

gertjanvanzwieten:
Commit 1: make the arg_axes argument of partial_derivative optional, defaulting to all axes.
Commit 2: bugfix; result could be returned as an integer and subsequently crash on missing the .shape attribute.

Followup discussion on commit 1: is there any conceivable use case where we want to specify a subset of axes within a function argument, i.e. setting arg_axes to anything else than the newly suggested default? If not then I'd favour simplifying things a little by dropping the shape argument in _derivative. For 'localcoords' we can easily introduce a token object with a .shape attribute.

### DIFF
--- a/nutils/function.py
+++ b/nutils/function.py
@@ -632,8 +632,8 @@ class Orientation( ArrayFunc ):
   def _opposite( self ):
     return Orientation( self.ndims, 1-self.side )
 
-  def _derivative( self, var, shape, seen ):
-    return _zeros( shape )
+  def _derivative( self, var, axes, seen ):
+    return _zeros( _taketuple(var.shape,axes) )
 
 class Align( ArrayFunc ):
   'align axes'
@@ -688,8 +688,8 @@ class Align( ArrayFunc ):
     trans = [ ax - (ax>axis) for ax in self.axes if ax != axis ]
     return align( func, trans, self.ndim-1 )
 
-  def _derivative( self, var, shape, seen ):
-    return align( derivative( self.func, var, shape, seen ), self.axes+tuple(range(self.ndim, self.ndim+len(shape))), self.ndim+len(shape) )
+  def _derivative( self, var, axes, seen ):
+    return align( derivative( self.func, var, axes, seen ), self.axes+tuple(range(self.ndim, self.ndim+len(axes))), self.ndim+len(axes) )
 
   def _multiply( self, other ):
     if not _isfunc(other) and len(self.axes) == other.ndim:
@@ -733,8 +733,8 @@ class Get( ArrayFunc ):
     assert arr.ndim == self.ndim+2
     return arr[ self.item_shiftright ]
 
-  def _derivative( self, var, shape, seen ):
-    f = derivative( self.func, var, shape, seen )
+  def _derivative( self, var, axes, seen ):
+    f = derivative( self.func, var, axes, seen )
     return get( f, self.axis, self.item )
 
   def _get( self, i, item ):
@@ -762,10 +762,10 @@ class Product( ArrayFunc ):
     assert arr.ndim == self.ndim+2
     return numpy.product( arr, axis=-1 )
 
-  def _derivative( self, var, shape, seen ):
-    grad = derivative( self.func, var, shape, seen )
+  def _derivative( self, var, axes, seen ):
+    grad = derivative( self.func, var, axes, seen )
     funcs = stack( [ util.product( self.func[...,j] for j in range(self.func.shape[-1]) if i != j ) for i in range( self.func.shape[-1] ) ], axis=-1 )
-    return ( grad * funcs[(...,)+(_,)*len(shape)] ).sum( self.ndim )
+    return ( grad * funcs[(...,)+(_,)*len(axes)] ).sum( self.ndim )
 
     ## this is a cleaner form, but is invalid if self.func contains zero values:
     #ext = (...,)+(_,)*len(shape)
@@ -813,8 +813,8 @@ class Transform( ArrayFunc ):
     assert matrix.shape == (self.todims,self.fromdims)
     return matrix.astype( float )[_]
 
-  def _derivative( self, var, shape, seen ):
-    return _zeros( self.shape + shape )
+  def _derivative( self, var, axes, seen ):
+    return _zeros( self.shape+_taketuple(var.shape,axes) )
 
   def _opposite( self ):
     return Transform( self.todims, self.fromdims, 1-self.side )
@@ -859,14 +859,14 @@ class Function( ArrayFunc ):
   def _opposite( self ):
     return Function( self.ndims, self.stdmap, self.igrad, self.shape[0], 1-self.side )
 
-  def _derivative( self, var, shape, seen ):
-    if var == 'localcoords':
-      ndims, = shape
+  def _derivative( self, var, axes, seen ):
+    if isinstance( var, LocalCoords ):
+      ndims, = var.shape
       grad = Function( self.ndims, self.stdmap, self.igrad+1, self.shape[0], self.side )
       return grad if ndims == self.ndims \
         else dot( grad[...,_], Transform( self.ndims, ndims, self.side ), axes=-2 )
     else:
-      return _zeros( self.shape+shape )
+      return _zeros( self.shape+_taketuple(var.shape,axes) )
 
   def _take( self, indices, axis ):
     if axis != 0:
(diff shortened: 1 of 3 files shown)


nutils/function.py
```diff
@@ -3052,24 +3053,17 @@ def wrapper( *args, **kwargs ):
...
-      orig = ba.kwargs[arg_key]
-      var = DerivativeHelper( orig.shape, arg_axes )
-      kwargs[arg_key] = var
-    else:
-      raise ValueError( 'argument not found: {!r}'.format( arg_key ) )
+    args = list(ba.args)
+    kwargs = dict(ba.kwargs)
+    keyargs = args if isinstance(arg_key,int) else kwargs
+    orig = keyargs[arg_key]
+    orig_axes = arg_axes if arg_axes is not None else tuple(range(orig.ndim))
```
joostvanzwieten: Small remark. I guess I would change `orig_axes = ...` into `arg_axes = ...`, e.g.

    if arg_axes is None:
      arg_axes = tuple(range(orig.ndim))

Not that `arg_axes` is well chosen, but `orig_axes` is just another name for the same thing. Then, you could revert the change on the last line.

nutils/function.py
```diff
@@ -3029,7 +3029,8 @@ def partial_derivative( func, arg_key, arg_axes ):
 
   if not isinstance( arg_key, (int, str) ):
     raise ValueError( 'arg_key: expected an int or str, got {!r}'.format( arg_key ) )
-  arg_axes = tuple(arg_axes)
+  if arg_axes is not None:
+    arg_axes = tuple(arg_axes) # evaluate iterator
```
joostvanzwieten: You could drop this statement and move `tuple` to where you create `orig_axes`, but I'd actually like that an error is raised here before anything has happened, if something invalid is passed to `arg_axes`. Eases debugging.

nutils/function.py
gertjanvanzwieten: I didn't try but I don't think that's possible; redefining arg_axes would make it local to wrapper so that would make the original argument unreachable.

joostvanzwieten:
I have the following in mind regarding the derivatives on not all axes: In DG some stabilisation schemes, numerical fluxes or whatever you want to call them, require integrating something involving `max(0, eig(f.J(u)))` over a path connecting two states at an element edge. `f.J` is in this case the jacobian of a flux function `f` w.r.t. the first argument limited to the axes of conserved quantities. `f` would be 'pointwise' on the other axes of `u`. The path integral could then be written as `(weights * max(0, eig(f.J(path(u, function.opposite(u)))))).sum(-1)` where `path` generates a set of integration points.

Long story short: I would keep this functionality. But I do like the idea of introducing a class `LocalCoords` and adding not `shape` but a `dshape` attribute to both the `DerivativeHelper` and the `LocalCoords` as a replacement for the `shape` argument to `_derivative`.

nutils/function.py
joostvanzwieten: Yes, you're right. Then everything seems OK.

gertjanvanzwieten:
Ok.. you lost me in the sixth bracket.
Seriously, as far as I understand the example and `u` contains the
conserved quantities I don't quite see why that could not always be made a
function of just those conserved quantities, and whatever else is required
in the composition, so that these conserved quantities can be targeted
directly in the differentiation. But if that's the case then there is
clearly merit to the current form so let's just leave it as it is.
I'm not too fond if the `dshape` suggestion because that's a departure from
the ArrayFunc form. If we'd consider a LocalCoords object I favour giving
it a `.shape` and change the `shape` argument in `_derivative` to `axes`,
equal to the `arg_axes` argument in `partial_derivative`.

On Tue, Mar 1, 2016 at 10:41 PM, Joost van Zwieten <notifications@github.com

joostvanzwieten:
For the `DerivativeHelper` it would be an extension, not a departure. Just add the following line to `DerivativeHelper.__init__`:

    self.dshape = tuple(self.shape[i] for i in self._axes)

The `LocalCoords` object would not be used as an `ArrayFunc`, right?

joostvanzwieten:
And probably `dshape` should have an underscore as it is only used internally.

gertjanvanzwieten:
It would be an extension but one that prevents the use of other ArrayFuncs
such as Guard. Not something we'd necessarily be interested in using, but I
like the current concept that var is really just any ArrayFunc that we can
retrace in the tree.
Wouldn't `axes` instead of `shape` also be slightly be more informative? It
tells you exactly which axes of `var` are being used and in which order, as
opposed to just the resulting shape.

On Tue, Mar 1, 2016 at 11:54 PM, Joost van Zwieten <notifications@github.com

joostvanzwieten:
> It would be an extension but one that prevents the use of other ArrayFuncs such as Guard. Not something we'd necessarily be interested in using, but I like the current concept that var is really just any ArrayFunc that we can retrace in the tree.

But using any `ArrayFunc` is pointless unless this `ArrayFunc` implements a `_derivative` method like `DerivativeHelper._derivative`. Actually, I think we should explicitly disallow passing any `ArrayFunc` to `function.derivative`, because the result is *not* the derivative with respect to this object, but zero.

> Wouldn't `axes` instead of `shape` also be slightly be more informative? It tells you exactly which axes of `var` are being used and in which order, as opposed to just the resulting shape.

This is only relevant to `DerivativeHelper._derivative`. However, this would indeed allow any `ArrayFunc` to be used, so if we are allowing this, changing `shape` to `axes` is fine by me.

gertjanvanzwieten:
Commit 3 adds unit tests in function (and fixes some issues in diagonalize and takediag for len(shape)>1).
Commit 4 changes the `shape` argument in `_derivative` to `axes`.

I don't think we should disallow using ArrayFuncs in derivative; in fact I believe that instead of the LocalCoords object introduced in commit 4 we can use the already existing ElemFunc* which indeed evaluates to the local coordinates and which derivative correctly returns eye if var==LocalCoords (of the same dimension).

(*and seize to opportunity to rename it to something sane, like, well, LocalCoords)

nutils/function.py
```diff
@@ -2405,15 +2407,21 @@ def evalf( self ):
...
-      result = 1
+      assert axes == self._axes
+      result = numpy.array(1)
       for i, axis in enumerate( self._axes ):
-        result *= align( eye( self.shape[axis] ), ( axis, self.ndim+i ), self.ndim+len(self._axes) )
+        result = result * align( eye( self.shape[axis] ), ( axis, self.ndim+i ), self.ndim+len(self._axes) )
       return result
     else:
-      return _zeros( self.shape + shape )
+      return _zeros( self.shape+self._derivshape )
```
joostvanzwieten: Shouldn't this be `_zeros( self.shape + _taketuple(var.shape, axes) )`?

nutils/function.py
```diff
@@ -2397,6 +2398,7 @@ class DerivativeHelper( ArrayFunc ):
   def __init__( self, shape, axes ):
     self._axes = tuple(axes)
     assert all(0 <= axis < len(shape) for axis in self._axes)
+    self._derivshape = _taketuple( shape, self._axes )
```
joostvanzwieten: Is the attribute `_derivshape` going to used somewhere?

gertjanvanzwieten: Check, will change it. Indeed then it's not.

nutils/function.py
```diff
@@ -2397,6 +2398,7 @@ class DerivativeHelper( ArrayFunc ):
   def __init__( self, shape, axes ):
     self._axes = tuple(axes)
     assert all(0 <= axis < len(shape) for axis in self._axes)
```
joostvanzwieten: Since `shape` has been replaced by `axes` in `ArrayFunc._derivative` I'm not sure if it's necessary to store the same `axes` as an attribute here.

gertjanvanzwieten: Good point. Fixing.

joostvanzwieten:
There is a hypothetical issue. If you are taking two nested derivatives to two different variables with same shape, the result might be wrong. Consider the following example:

    a = DerivativeHelper( [2] )
    b = DerivativeHelper( [2] )
    derivative( derivative( a * b, a, [] ), b, [] )

the result should be 1, not 2. This is caused by `a` and `b` being identical. I guess the only solution is to add a unique token to `DerivativeHelper`. And to hide the ugliness, I propose the following function:

    def derivative_helper( shape ):
      return DerivativeHelper( shape, token=object() )

This actually uncovered a bug in `DerivativeHelper._derivative`. The initial result should not be `numpy.array(1)`, but a 1 with shape `self.shape+_taketuple(var.shape,axes)`. See [45031aa] for a fix.

[45031aa]: https://github.com/joostvanzwieten/nutils/commit/45031aaef5694ffffdc7a556c99fb10e4ca4151c

gertjanvanzwieten:
Commit 5 removes the ad hoc LocalCoords object and renames the pre-existing ElemFunc to LocalCoords.
Commit 6 introduces a DerivativeTargetBase class for error checking and renames DerivativeHelper to DerivativeTarget.

Suggesting to pull as is and leave 45031aa for a follow-up.

joostvanzwieten:
Looks good!

PR #99, opened 2016-03-01
Spsolve

gertjanvanzwieten:
Fixes failing unit tests on python 3.3 by using a direct solver. It is unclear what exactly causes the issues but differences arise in the solution vector, and disappear when switching to a direct solver or setting a sharper tolerance. The difference may not be due to different python versions but due to difference scipy versions, 0.16.0 vs 0.17.0.

### DIFF
--- a/examples/adaptivity
+++ b/examples/adaptivity
@@ -107,7 +107,7 @@ def main( degree=1, solvetol=1e-10, circle=False, uniform=False, basistype='std'
 
 def unittest():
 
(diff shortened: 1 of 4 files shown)


gertjanvanzwieten:
It turns out scipy's spsolve is affected as well, and that the problem is sensitive to the permutation setting. Solving by disabling row/column permutations in tiny systems until we find a better solution.

PR #128, opened 2016-08-05
rewrite ortho_complement using svd

gertjanvanzwieten:


nutils/numeric.py
```diff
@@ -332,33 +332,14 @@ def isint( a ):
...
-
-  nextalpha = numpy.max( numpy.linalg.norm( X, axis=1 ) )
-  assert nextalpha < tol, '{} > {}'.format( nextalpha, tol )
-  return Y
+def ortho_complement( A ):
+  '''return orthogonal complement to non-square matrix A'''
+
+  m, n = A.shape
+  assert n < m
+  numpy.linalg.svd( A )
```
timovanopstal: typo? computing once and trashing the output, and computing again it seems

gertjanvanzwieten: good catch! fixing, thanks

PR #186, opened 2017-01-25
new command line interface, additions and fixes

gertjanvanzwieten:


nutils/util.py
```diff
@@ -419,7 +419,7 @@ def run( *functions ):
       assert arg in properties, 'invalid argument %r' % arg
       properties[arg] = val
 
-  missing = [ arg for arg, val in kwargs.items() if val is parameter.empty ]
+  missing = [ arg for arg, val in kwargs.items() if val is inspect._empty ]
```
joostvanzwieten: `inspect._empty` is undocumented. I think this should be [`inspect.Parameter.empty`](https://docs.python.org/3/library/inspect.html#inspect.Parameter.empty).

gertjanvanzwieten: fixed dc9c44f

PR #188, opened 2017-02-01
Minor additions

gertjanvanzwieten:


gertjanvanzwieten:
@timovanopstal My personal take on the new colormap is that I dislike it, and that a scientifically interesting design goal resulted in producing just different shades of blue - and barely different at that. But indeed that last commit is controversial, also within Evalf HQ. My rational was that we are at liberty to change the default behaviour of functions we added ourselves (the .mesh method), but of course the cmap argument was always present and I am sensitive to the argument that the 'PyPlot' interface should not aim to change matplotlib's default behaviour. All that is to day that I will remove the commit. Thanks for the input!

timovanopstal:
Thanks for clarifying... and reverting ;-)

PR #190, opened 2017-02-11
Minor changes

gertjanvanzwieten:


nutils/log.py
```diff
@@ -487,27 +478,44 @@ def range( title, *args ):
...
+    length = _len(iterable)
+  log = _getlog()
+  index = 0
+  for item in iterable:
+    text = '{} {}'.format( title, index )
+    if length:
+      text += ' ({:.0f}%)'.format( (index+.5) * 100. / length )
+    with log.context( text ):
+      yield item
+    index += 1
```
joostvanzwieten: Instead of

    index = 0
    for item in iterable:
        ....
        index += 1

you could write

    for index, item in enumerate( iterable ):
        ...

nutils/log.py
```diff
@@ -487,27 +478,44 @@ def range( title, *args ):
   '''Progress logger identical to built in range'''
 
   items = _range( *args )
-  return _logiter( title, _iter(items), length=len(items), useitem=True )
+  log = _getlog()
+  for item in items:
+    with log.context( '{} {} ({:.0f}%)'.format( title, item, (item+.5) * 100. / len(items) ) ):
+      yield item
```
joostvanzwieten: Does it make sense to add `50 / len(items)` to the percentage? The work for this iteration is done *after* logging the context.

nutils/log.py
gertjanvanzwieten: see new code in 9ffcde2

nutils/log.py
gertjanvanzwieten: Agreed, if you mean to suggest adding 100 instead of 50. Actually the reason I started touching this code (and remove the `useitem` argument) was to log the item generation inside the context, in which case we should add none at all. See new commit 9ffcde2 where I implement this idea.

PR #198, opened 2017-03-20
Extra bases

joostvanzwieten:


nutils/element.py
```diff
@@ -1482,6 +1482,16 @@ def bernstein_poly( cls, degree ):
     return revpoly[::-1]
 
   @classmethod
+  def lagrange_poly( cls, degree ):
```
joostvanzwieten: This function generates Lagrange polynomials having an interpolation property on an equidistant grid. Shouldn't 'equidistant' be part of the name? 'lagrange' seems a bit ambiguous.

gertjanvanzwieten: Agreed.

joostvanzwieten:
With #233 this has to be re-implemented.

PR #205, opened 2017-06-29
Some mesh.multipatch fixes

TheBB:
* First commit: It was incorrectly raising errors on `(npatches,2,...,2)`-type input. Also added a check that the last dimension has an even size, for good measure.

* Second commit: Make the `None` key for default numbers of elements optional.

### DIFF
--- a/nutils/mesh.py
+++ b/nutils/mesh.py
@@ -217,7 +217,9 @@ def multipatch( patches, nelems, patchverts=None, name='multipatch' ):
   patches = numpy.array( patches )
   if patches.dtype != int:
     raise ValueError( '`patches` should be an array of ints.' )
-  if len( patches.shape ) != 2 or patches.ndim > 2 and patches.shape[1:] != (2,)*(patches.ndim-1):
+  if patches.ndim < 2 or patches.ndim == 2 and patches.shape[-1] % 2 != 0:
+    raise ValueError( '`patches` should be an array with shape (npatches,2,...,2) or (npatches,2*ndims).' )
+  elif patches.ndim > 2 and patches.shape[1:] != (2,) * (patches.ndim - 1):
     raise ValueError( '`patches` should be an array with shape (npatches,2,...,2) or (npatches,2*ndims).' )
   patches = patches.reshape(patches.shape[0], -1)
 


TheBB:
CI failure seems unrelated. I guess Travis has a newer numpy.

nutils/mesh.py
```diff
@@ -217,7 +217,9 @@ def multipatch( patches, nelems, patchverts=None, name='multipatch' ):
   patches = numpy.array( patches )
   if patches.dtype != int:
     raise ValueError( '`patches` should be an array of ints.' )
-  if len( patches.shape ) != 2 or patches.ndim > 2 and patches.shape[1:] != (2,)*(patches.ndim-1):
+  if patches.ndim == 2 and patches.shape[-1] % 2 != 0:
+    raise ValueError( '`patches` should be an array with shape (npatches,2,...,2) or (npatches,2*ndims).' )
+  elif patches.ndim > 2 and patches.shape[1:] != (2,) * (patches.ndim - 1):
     raise ValueError( '`patches` should be an array with shape (npatches,2,...,2) or (npatches,2*ndims).' )
```
joostvanzwieten: To be complete, could you extend the tests to `patches` with `ndim < 2`?

TheBB:
Updated :+1:

nutils/mesh.py
```diff
@@ -267,7 +269,7 @@ def multipatch( patches, nelems, patchverts=None, name='multipatch' ):
         sideverts = frozenset(shaped_patch[side])
         if sideverts in nelems:
           nelems_sides.append(nelems[sideverts])
-        else:
+        elif None in nelems:
           nelems_sides.append(nelems[None])
       if len(set(nelems_sides)) != 1:
         raise ValueError( 'duplicate number of elements specified for patch {} in dimension {}'.format( i, dim ) )
```
joostvanzwieten: This was intended behavior. Every patch boundary should either be mentioned in `nelems` explicitly, or the boundary uses the default `nelems[None]`. I admit that if the default is missing the error message — currently `KeyError` — could be improved. I'm not really in favor of this change, but let's see what others think of this.

Anyhow, I believe this introduces the following problem. Suppose we are creating two patches with one shared boundary, we specify different numbers of elements at the opposing boundaries and leave the shared boundary undefined:

    # connectivity:  3─────4─────5
    #                │ 1x2 │ 1x3 │
    #                0─────1─────2
    mesh.multipatch(patches=[ [0,3,1,4], [1,4,2,5] ], nelems={(0,3):2, (2,5):3, (0,1):1, (3,4):1, (1,2):1, (4,5):1})

With the proposed changes this won't raise an exception anymore, but will cause issues further down the road when we want to create a C^0 patch-continuous basis or integrate over the patch interface `(1,4)`: the boundaries of the patches do not match. If we do want the loose behavior, we should add appropriate tests in `mesh.multipatch()`.

TheBB:
I removed the second commit. Turned out I had made some faulty assumptions about how it worked.

PR #214, opened 2017-08-13
Improve NumpyMatrix arithmetic

TheBB:
- Implement missing __radd__
- Support other types

Same implementation as ScipyMatrix

### DIFF
--- a/nutils/matrix.py
+++ b/nutils/matrix.py
@@ -192,9 +192,10 @@ class NumpyMatrix( Matrix ):
 
   matvec = lambda self, vec: numpy.dot( self.core, vec )
   toarray = lambda self: self.core
-  __add__ = lambda self, other: NumpyMatrix( self.core + other.toarray() )
-  __sub__ = lambda self, other: NumpyMatrix( self.core - other.toarray() )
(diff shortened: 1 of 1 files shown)


gertjanvanzwieten:
Though adding support for Numpy arrays seems reasonable enough, I'm not sure if we should place nutils.Matrix objects at equal footing with Numpy and Scipy matrices. The idea for Matrix objects is that they provide an abstraction layer for several different matrix backends (currently only Numpy and Scipy, but with aims of extending to Trilinios and others) in order to provide a consistent interface to the user. Arrays can be explicitly brought into (NumpyArray, ScipyArray) and out of (.toarray, .toscipy) this world if necessary, but by making Arrays and non-Arrays directly interoperable the distinction might get a little muddled. That is not to suggest the current matrix module is perfect -far from- or in any way set in stone, but I do feel that the change proposed here is a step in a different direction than the one envisioned.

TheBB:
While I don't know about all that, it's just inconsistent with ScipyMatrix.

```python
# Works
sum(domain.integrate(list_of_functions, ...))
# Doesn't work, 0 + NumpyMatrix is illegal
sum(domain.integrate(list_of_functions, force_dense=True))
```

PR #216, opened 2017-08-25
Improved midpoint method and 3D gmsh reader

CVerhoosel:


joostvanzwieten:
Clemens, could you perhaps remove the last commit (Merge branch master into midpoint) and instead rebase midpoint onto master? The following commands are sufficient:

    # Backup the original midpoint branch.
    git branch -m midpoint midpoint-old
    # Create and checkout a new midpoint branch sans merge commit.
    git checkout -b midpoint midpoint-old^
    # Rebase midpoint onto master.
    git rebase master

You should then force-push the branch to github (add flag `-f` in `git push ...`).

gertjanvanzwieten:
Looks great! Two small suggestions:
- please return _children_vertices as a tuple (because it's a cached property it's good practice to make it immutable)
- please move the expression from _from_multi_index inside the list comprehension inside TetrahedronReference.child_divide; this is the only place from where it's called so the static method is unnecessary complexity.

PR #243, opened 2017-11-09
Replace top-level lambdas with defs (fixes #213)

### LINKED ISSUE #213: Prevalence of top-level lambdas makes pickling difficult
In `function.py` there's about 50 functions defined as top-level lambdas.

```python
zeros_like = lambda arr: zeros( arr.shape, _jointdtype(arr.dtype) )
```

This is nice for brevity but functionally not the same as a regular *def*. In particular, lambdas can't be pickled, while module-level *def*s can.

For example, this fails.

```python
domain, geom = mesh.rectilinear([[-1, 1]])
h = function.heaviside(geom)  # the resulting object references the function.zeros_like object
pickle.dumps(h)
```

Rewriting `zeros_like` in `def` form makes it work.

TheBB:
Fixes #213 

Note, I've overridden the default argument in `cache.replace` (which is a lambda) in three cases, using it as a decorator instead.

### DIFF
--- a/nutils/cache.py
+++ b/nutils/cache.py
@@ -248,7 +248,7 @@ class Tuple( object ):
       self.__items[i] = item = self.__getitem( self.__start + i * self.__stride )
     return item
 
-def replace(func=lambda obj: None, initcache={}):
+def replace(func=None, initcache={}):
   '''decorator for deep object replacement
 
   Generates a replacement method for Immutable objects. Replacements can be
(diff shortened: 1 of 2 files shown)


TheBB:
Actually the second commit can be dropped. It doesn't help since the return value is still a dynamically created function without a module-level path. I don't need it for my code though.

gertjanvanzwieten:
merged with minor modification in 2nd commit

PR #246, opened 2017-11-20
pyplot: add optionally filled contours

TheBB:
`tricontour` and `tricontourf` have identical interfaces for all practical purposes. It seems reasonable to serve both of them from the same entry point.

### DIFF
--- a/nutils/plot.py
+++ b/nutils/plot.py
@@ -190,7 +190,7 @@ class PyPlot( BasePlot ):
       values[mask] = numpy.mean( values[~mask] ) #Set masked values to average to not affect color range
     return self._pyplot.tripcolor( tri, values, **kwargs )
 
-  def tricontour( self, tri, values, every=None, levels=None, mergetol=0, **kwargs ):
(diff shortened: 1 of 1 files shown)


gertjanvanzwieten:
Though I agree that it seems reasonable to serve both from the same entry point, perhaps more consistent with the existing framework of loosely wrapping matplotlib's Axes object would be to add tricountourf as a separate method, rather than adding the fill argument. Not sure if you considered and rejected this, but in case you agree could you modify the PR accordingly, for example by feeding the backend argument into a third method?

Some context:

The plot module, and specifically the PyPlot object, are still very much in search of a final form. In particular the present context manager (with plot.PyPlot..) seems to be a poor choice as it prevents the simultaneous construction of multiple images, and we would like to steer free from matplotlib's state machine (current image, current axes, etc) while still keeping access to matplotlib's api. Given this status there is no point worrying much about introducing things that we might break later, so additions like the one proposed can certainly go in without much scrutiny. The suggested modification is just personal preference.

TheBB:
Is this more like what you had in mind?

gertjanvanzwieten:
Thanks!

PR #281, opened 2018-02-14
introduce matrix backends, add MKL matrix, pardiso solver

gertjanvanzwieten:


nutils/numeric.py
```diff
@@ -507,4 +507,26 @@ def poly_pow(p, n):
     return poly_mul(q, p)
   return q
 
+def accumulate(data, index, shape):
+  '''accumulate scattered data in dense array.
+
+  Accumulates values from `data` in an array of shape `shape` at positions
+  `index`, equivalent with:
```
joostvanzwieten: I think you have to use double backticks in rst for [code literals](http://www.sphinx-doc.org/en/stable/rest.html#inline-markup).

gertjanvanzwieten: Fixed in d46b7. There will probably be more of those, will search through remaining commits tomorrow.

PR #282, opened 2018-02-15
fix ischeme vtk for point*tensor reference

joostvanzwieten:
The `getischeme_vtk` method of the `TensorReference` generates integration
points with incorrect order if `ref1` has dimension zero.  Such references
originate for example from the boundary in dimension zero of a structured
topology.  Since `ref1` has no contribution to the integration points anyhow,
this patch fixes this issue by forwarding `getischeme_vtk` to `ref2`.

### DIFF
--- a/nutils/element.py
+++ b/nutils/element.py
@@ -820,7 +820,10 @@ class TensorReference(Reference):
     return '{}*{}'.format(self.ref1, self.ref2)
 
   def getischeme_vtk(self):
-    if self.ref1.ndims == self.ref2.ndims == 1:
+    if self.ref1.ndims == 0:
+      assert self.ref1.nverts == 1
+      points, weights = self.ref2.getischeme_vtk()
+    elif self.ref1.ndims == self.ref2.ndims == 1:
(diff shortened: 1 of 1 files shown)


gertjanvanzwieten:
Won't this issue go away when we switch to pixels, voxels? Then instead of adding a special case we can remove one.

joostvanzwieten:
According to the VTK docs the edges of [pixel] and [voxel] elements are aligned with the coordinate axes, so I doubt we can use these element types. Nevertheless, we can switch to ischeme `vertex` and apply the necessary reordering for [quad] and [hexahedron] in the VTK plot routine, but that requires a bit more work which I'd like to postpone for a bit.

[pixel]: https://www.vtk.org/doc/release/7.1/html/classvtkPixel.html#details
[voxel]: https://www.vtk.org/doc/release/7.1/html/classvtkVoxel.html#details
[quad]: https://www.vtk.org/doc/release/7.1/html/classvtkQuad.html#details
[hexahedron]: https://www.vtk.org/doc/release/7.1/html/classvtkHexahedron.html#details

PR #287, opened 2018-03-03
Fixes #286: $-character is parsed correctly for expressions

### LINKED ISSUE #286: δ and $ are not interchangeable in expression syntax
If I run `examples/elasticity.py` with the sole change of swapping out the δ-symbol with a $-symbol, then I get the following exception

```
Traceback (most recent call last):
  File "/home/kjetijo/3rdPartyLibs/nutils/nutils/expression.py", line 516, in wrapper
    return f(self, *args, **kwargs)
  File "/home/kjetijo/3rdPartyLibs/nutils/nutils/expression.py", line 965, in tokenize
    raise _IntermediateError('Unknown symbol: {!r}.'.format(self.expression[pos]), at=pos)
nutils.expression._IntermediateError: Unknown symbol: '$'.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/kjetijo/3rdPartyLibs/nutils/nutils/cli.py", line 197, in call
    func(**kwargs)
  File "elasticity.py", line 29, in main
    ns.stress_ij = 'lmbda strain_kk $_ij + 2 mu strain_ij'
  File "/home/kjetijo/3rdPartyLibs/nutils/nutils/function.py", line 3566, in __setattr__
    ast, arg_shapes = expression.parse(value, variables=self._attributes, functions=self._functions_nargs, indices=indices, arg_shapes=self._arg_shapes, default_geometry_name=self.default_geometry_name)
  File "/home/kjetijo/3rdPartyLibs/nutils/nutils/expression.py", line 1193, in parse
    parser.tokenize()
  File "/home/kjetijo/3rdPartyLibs/nutils/nutils/expression.py", line 524, in wrapper
    raise ExpressionSyntaxError(e.msg + '\n' + self.expression + '\n' + ' '*at + '^'*count) from e
nutils.expression.ExpressionSyntaxError: Unknown symbol: '$'.
lmbda strain_kk $_ij + 2 mu strain_ij
                ^
```
```

VikingScientist:
With this fix, then I do manage to run the example file in question, but I'm not sure if I break other things.

Instructions on how to run the complete test suite would be welcome.

### DIFF
--- a/nutils/expression.py
+++ b/nutils/expression.py
@@ -906,11 +906,14 @@ class _ExpressionParser:
         tokens.append(_Token(self.expression[pos], self.expression[pos], pos))
         pos += 1
         continue
+      m = re.match(r'[%s|%s]'%tuple(self.eye_symbols), self.expression[pos:])
+      if m:
+        tokens.append(_Token('eye', m.group(0), pos))
+        pos += m.end()
+        continue
       m = re.match(r'[?]?[a-zA-Zα-ωΑ-Ω][a-zA-Zα-ωΑ-Ω0-9]*', self.expression[pos:])
       if m:
-        if m.group(0) in self.eye_symbols:
-          tokens.append(_Token('eye', m.group(0), pos))
-        elif m.group(0) in self.normal_symbols:
+        if m.group(0) in self.normal_symbols:
           tokens.append(_Token('normal', m.group(0), pos))
         else:
           tokens.append(_Token('variable', m.group(0), pos))
--- a/tests/test_expression.py
+++ b/tests/test_expression.py
@@ -366,8 +366,23 @@ class parse(TestCase):
 
   # EYE
 
-  def test_single_eye(self): self.assert_ast('a2_i δ_ij', 'j', ('sum', ('mul', ('append_axis', v._a2, _(2)), ('eye', _(2))), _(0)))
+  def test_single_eye(self):    self.assert_ast('a2_i δ_ij', 'j', ('sum', ('mul', ('append_axis', v._a2, _(2)), ('eye', _(2))), _(0)))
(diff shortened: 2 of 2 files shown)


gertjanvanzwieten:
The unit tests can be run using python3 setup.py test, or python3 -m unittest. In your current PR there seem to be two closing brackets missing for test_multiple_eye_v2.

joostvanzwieten:
I'd like to fix it in a slightly different way. This PR allows `$`-symbols in variable names. Thanks for spotting this!

VikingScientist:
closed the brackets in the test and changed the logic so  `$`-characters are no longer allowed in variable names.

joostvanzwieten:
If I'm not mistaken the proposed change would disallow variable names starting with `δ`. In fact, `δx` would be tokenized as `_Token('eye','δ',...), _Token('variable','x',...)` and the parser will fail at the eye token because it expects an underscore and indices. And you assume the length of `eye_symbols` to be 2, which to some extent defeats the purpose of not hard-coding the symbols in the `tokenize` method. I think the best way forward is to test for eye symbols (and similarly for normal symbols) and accept the token if it is not followed by something that looks like a variable. I'm happy to write an appropriate fix in the coming days.

PS1. We don't use the `%`-based string formatting in nutils anymore.

PS2. To be fail-safe, you should escape the `eye_symbols` with [`re.escape`](https://docs.python.org/3/library/re.html#re.escape).

VikingScientist:
>  I think the best way forward is to test for eye symbols (and similarly for normal symbols) and accept the token if it is not followed by something that looks like a variable. I'm happy to write an appropriate fix in the coming days.

Sounds great! I'll leave this PR here if you need it for reference, but just delete it if you don't need it or when you've submitted your fix to trunk.

joostvanzwieten:
Superseded by #290.

PR #366, opened 2018-09-28
fix loss of endianness for vector data

gertjanvanzwieten:


scdivi:
I confirm that the fix by Gertjan solves the problem in my case. Thanks!

tristandemont:
Likewise, I can confirm it has fixed the vector output in my case. Thank you!

Issue #284, opened 2018-02-21
arctan in namespace (functions.py)


scdivi:
Hi, 

I have found a small bug in **function.py** while using **arctan** in namespace. Figured out that it was not specified in the **_functions (line 3479)** and by including arctan=arctan, the bug was fixed. The git diff output below might be more clear.

diff --git a/nutils/function.py b/nutils/function.py
index abec666..3434208 100644
--- a/nutils/function.py
+++ b/nutils/function.py
@@ -3479,7 +3479,7 @@ class Namespace:

  _functions = dict(
    opposite=opposite, sin=sin, cos=cos, tan=tan, sinh=sinh, cosh=cosh,
 ---  tanh=tanh, arcsin=arcsin, arccos=arccos, arctan2=arctan2, arctanh=arctanh,
+++ tanh=tanh, arcsin=arcsin, arccos=arccos, **arctan = arctan**, arctan2=arctan2, arctanh=arctanh,
    exp=exp, abs=abs, ln=ln, log=ln, log2=log2, log10=log10, sqrt=sqrt,
    sign=sign,
  )

gertjanvanzwieten:
Thanks, we will fix it shortly! (or feel free to send a pull request)

PR #368, opened 2018-10-02
Added uniform integration scheme to Cones

scdivi:
Added uniform integration scheme to Cones #368


scdivi:
Additionally also closed the issue: arctan in namespace (functions.py) #284

gertjanvanzwieten:
split in #370 and #371

PR #380, opened 2018-11-28
Lazy references, transforms and bases

joostvanzwieten:


nutils/topology.py
```diff
@@ -1712,255 +1769,252 @@ def connectivity(self):
...
-        opptrans = basebtopo.elements[iedge].opposite + tail
-        belems.append(element.Element(edge.reference, edge.transform, opptrans))
-    return basebtopo.hierarchical(belems, precise=True)
+    bindices_per_level = []
+    blevel = None
+    for level, indices in zip(self.levels, self._indices_per_level):
+      blevel = basebtopo if blevel is None else blevel.refined
+      bindex = blevel.transforms.index
+      bindices = []
+      for trans, ref in zip(map(level.transforms.__getitem__, indices), map(level.references.__getitem__, indices)):
```
gertjanvanzwieten: Don't we loose performance here by looping over all elements instead of "edgepool"?  I imagine that concept can still be applied per level to reduce indices since we already have blevel.

nutils/transform.py
```diff
@@ -538,6 +499,7 @@ class Identifier(TransformItem):
 
   @types.apply_annotations
   def __init__(self, ndims:int, *args):
+    assert all(not isinstance(arg, Identifier) for arg in args)
```
gertjanvanzwieten: does this serve a purpose? looks like leftover debug code

nutils/element.py
```diff
@@ -253,20 +253,20 @@ def vertex_cover(self, ctransforms, maxrefine):
     npoints = self.nvertices_by_level(maxrefine)
     allindices = numpy.arange(npoints)
     if len(ctransforms) == 1:
-      assert not ctransforms[0]
+      assert not next(iter(ctransforms))
       return ((), self.getpoints('vertex', maxrefine).coords, allindices),
     if maxrefine == 0:
       raise Exception('maxrefine is too low')
-    cbins = [[] for ichild in range(self.nchildren)]
+    cbins = [set() for ichild in range(self.nchildren)]
```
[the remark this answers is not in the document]
joostvanzwieten: Quick benchmark:

```
In [6]: %%timeit a = []
   ...: for i in range(10000):
   ...:   a.append(i)
   ...: b = frozenset(a)
   ...: 
100 loops, best of 5: 11.1 ms per loop

In [7]: %%timeit a = set()
   ...: for i in range(10000):
   ...:   a.add(i)
   ...: b = frozenset(a)
   ...: 
1000 loops, best of 5: 982 µs per loop
```

nutils/topology.py
joostvanzwieten: In `examples/adaptivity.py` creating the boundary is not so expensive. Not sure if this is worth the effort at this point.

nutils/element.py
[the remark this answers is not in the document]
gertjanvanzwieten: Mindblowing.

nutils/topology.py
gertjanvanzwieten: ok

nutils/topology.py
```diff
@@ -1712,255 +1769,252 @@ def connectivity(self):
...
-      for belem in trimmed:
-        trimmedbrefs[self.newboundary.edict[belem.transform]] = belem.reference
+      trimmedbrefs = [ref.empty for ref in self.newboundary.references]
+      for ref, trans in zip(trimmedreferences, trimmedtransforms):
+        trimmedbrefs[self.newboundary.transforms.index(trans)] = ref
       trimboundary = SubsetTopology(self.newboundary, trimmedbrefs)
     else:
-      trimboundary = OrientedGroupsTopology(self.basetopo.interfaces, trimmed)
-    return UnionTopology([trimboundary, origboundary], names=[self.newboundary] if isinstance(self.newboundary,str) else [])
+      trimboundary = OrientedGroupsTopology(self.basetopo.interfaces, trimmedreferences, trimmedtransforms, trimmedopposites)
```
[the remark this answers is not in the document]
joostvanzwieten: Then the unittest `tests.test_mesh.gmshrect:2d.test_iface` is esoteric!

PR #382, opened 2018-12-11
enable pickling of function.Namespace instances

timovanopstal:
I have no clue if this would be of interest to others. If so, I could accompany this with tests and incorporate any suggestions..

### DIFF
--- a/nutils/function.py
+++ b/nutils/function.py
@@ -3765,7 +3765,7 @@ class Namespace:
       The name of the default geometry.  See argument with the same name.
   '''
 
-  __slots__ = '_attributes', '_arg_shapes', 'arg_shapes', 'default_geometry_name'
+  __slots__ = '_attributes', '_arg_shapes', 'default_geometry_name'
 
   _re_assign = re.compile('^([a-zA-Zα-ωΑ-Ω][a-zA-Zα-ωΑ-Ω0-9]*)(_[a-z]+)?$')
 
@@ -3785,9 +3785,21 @@ class Namespace:
       raise ValueError('default_geometry_name: Invalid variable name: {!r}.'.format(default_geometry_name))
     super().__setattr__('_attributes', {})
     super().__setattr__('_arg_shapes', {})
-    super().__setattr__('arg_shapes', builtin_types.MappingProxyType(self._arg_shapes))
     super().__setattr__('default_geometry_name', default_geometry_name)
     super().__init__()
+
+  def __getstate__(self):
+    'Pickle instructions'
+    attrs = '_arg_shapes', '_attributes', 'default_geometry_name'
+    return {k: getattr(self, k) for k in attrs}
+
+  def __setstate__(self, d):
+    'Unpickle instructions'
+    for k, v in d.items(): super().__setattr__(k, v)
+
+  @property
+  def arg_shapes(self):
+    return builtin_types.MappingProxyType(self._arg_shapes)
 
   @property
   def default_geometry(self):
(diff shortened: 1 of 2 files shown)


nutils/function.py
```diff
@@ -3789,6 +3789,22 @@ def __init__(self, *, default_geometry_name='x'):
...
+      try:
+        return attr, getattr(self, attr)
+      except AttributeError as e:
+        import inspect
+        log.warning('{}.{}.{} failed to add {}'.format(cls.__module__,
+          cls.__name__, inspect.stack()[2].function, attr))
+        return attr, None
+    attrs = '_arg_shapes', '_attributes', 'default_geometry_name'
+    return {k: v for k, v in map(catchgetattr, attrs) if v is not None}
+
```
joostvanzwieten: Is `catchgetattr` really necessary? The attributes to be pickled are set in `__init__`, so I don't see why `getattr` might fail.

timovanopstal: Absolutely right.

nutils/function.py
```diff
@@ -3789,6 +3789,15 @@ def __init__(self, *, default_geometry_name='x'):
...
 
+  def __getstate__(self):
+    'Pickle instructions'
+    attrs = '_arg_shapes', '_attributes', 'default_geometry_name'
+    return {k: getattr(self, k) for k in attrs}
+
+  def __setstate__(self, d):
+    'Unpickle instructions'
+    for k, v in d.items(): super().__setattr__(k, v)
+
```
joostvanzwieten: Since pickle bypasses `__init__`  when unpickling an object with `__getstate__`, `__setstate__`, the attribute `arg_shapes` is undefined for unpickled `Namespace` objects. Could you fix this by converting the `arg_shapes` attribute into a property?

tests/test_function.py
```diff
@@ -665,6 +665,21 @@ def test_replace(self):
...
+  def test_pickle(self):
+    orig = function.Namespace()
+    domain, geom = mesh.unitsquare(2, 'square')
+    orig.x = geom
+    orig.v = domain.basis('std', 1)
+    orig.u = 'v_n ?lhs_n'
+    orig.f = 'cosh(x_0)'
+    pickled = pickle.loads(pickle.dumps(orig))
+    for attr in ('x', 'v', 'u', 'f'):
+      self.assertEqual(getattr(pickled, attr), getattr(orig, attr))
```
joostvanzwieten: Can you add a test for the `arg_shapes` attribute? For example

    self.assertEqual(pickled.arg_shapes['lhs'], orig.arg_shapes['lhs'])

tests/test_function.py
```diff
@@ -665,6 +665,21 @@ def test_replace(self):
...
+
+  def test_pickle_default_geometry_name(self):
+    orig = function.Namespace()
+    pickled = pickle.loads(pickle.dumps(orig))
+    self.assertEqual(pickled.default_geometry_name, orig.default_geometry_name)
```
joostvanzwieten: Can you pick a non-default geometry name here? For example

    orig = function.Namespace(default_geometry_name='foo')

timovanopstal: Thanks for the review, cf forthcoming commit

joostvanzwieten: Could you move the `default_geometry_name=foo` bit to the `test_pickle_default_geometry_name` function? In the other function the value of `pickle.default_geometry_name` is not checked. You can revert the `default_geometry_name` changes in `test_pickle`. Sorry if this was not clear.

timovanopstal:
Sorry, that makes more sense of course. See the forthcoming.

PR #386, opened 2018-12-14
treat removedofs in stdbasis equal to splinebasis

gertjanvanzwieten:


scdivi:
I can confirm that it works for me.

PR #395, opened 2019-01-23
fix nested 'withgroups' topologies

gertjanvanzwieten:


nutils/topology.py
```diff
@@ -845,7 +833,16 @@ def interfaces(self):
...
+    ptopos = []
+    pnames = []
+    topo = self
+    while isinstance(topo, WithGroupsTopology):
+      for pname, ptopo in topo.pgroups.items():
+        if pname not in pnames:
+          pnames.append(pname)
+          ptopos.append(ptopo)
+      topo = topo.basetopo
+    return DisjointUnionTopology(ptopos, pnames)
```
joostvanzwieten: Are the pgroups/ptopos always disjoint?

gertjanvanzwieten: In changing to UnionTopology I must have pushed a wrong commit causing github to close the PR. Opening a new one.

PR #413, opened 2019-05-21
Matrix

gertjanvanzwieten:


nutils/matrix.py
```diff
@@ -162,11 +162,13 @@ def solve(self, rhs=None, *, lhs0=None, constrain=None, rconstrain=None, solver=
...
     nrows, ncols = self.shape
+    if rhs is None:
+      rhs = numpy.zeros(nrows)
     if lhs0 is None:
-      x = numpy.zeros(ncols)
+      x = numpy.zeros((ncols,)+rhs.shape[1:])
     else:
       x = numpy.array(lhs0, dtype=float)
-      assert x.shape == (ncols,)
+      assert x.shape == (ncols,)+rhs.shape[1:]
```
joostvanzwieten: I'm not sure if this is of any practical use, but shouldn't we allow a single `lhs0` (`lhs0.shape` equals `(ncols,)`) with multiple right hand sides?

PR #416, opened 2019-05-29
add time target to thetamethod

joostvanzwieten:


nutils/solver.py
```diff
@@ -605,29 +605,36 @@ def __init__(self, target:types.strictstr, residual:sample.strictintegral, inert
     self.inertia = inertia
     self.theta = theta
     self.timestep = timestep
+    self.timetarget = timetarget or '_thetamethod_dummy'
```
gertjanvanzwieten: explain again why '_thetamethod_dummy' cannot simply be the default value for timetarget?

nutils/solver.py
```diff
@@ -605,29 +605,36 @@ def __init__(self, target:types.strictstr, residual:sample.strictintegral, inert
...
+      yield lhs, i
     while True:
-      lhs = self._step(lhs, self.timestep)
+      lhs = self._step(lhs, self.time0+i*self.timestep, self.timestep)
+      i += 1
+      yield lhs, i
+
+  def __iter__(self):
+    for lhs, i in super().__iter__():
       yield lhs
```
gertjanvanzwieten: Isn't it cleaner to have "i" be retrievable from the cacher, rather than caching the enumeration and then removing it? After all the cacher has this number, it just doesn't currently volunteer it. Not sure if that can be done in a backwards-compatible way though so that might argue against it.

gertjanvanzwieten:
Any idea why travis is failing? Whatever is going wrong it appears unrelated to the PR. Curious if it persists after a rebuild.

nutils/solver.py
joostvanzwieten: I would say `timetarget` is an optional argument in the sense of `typing.Optional`. In that case `None` is a valid value. `'_thetamethod_dummy'` is an implementation detail, which shouldn't be exposed to users in any way, e.g. by introspection or by reading the docs. As an alternative to the dummy value, we could also opt for conditional replace and argument statements on lines 612 and 620. Or we could choose a default `timetarget`, say `'t'`.

nutils/solver.py
joostvanzwieten: I think the only way we could do this is by adding an argument to the `resume` method, which breaks existing subclasses (outside Nutils). We could of course support a resume method with both one and two arguments (besides `self`) via introspection, and issue a deprecation warning in case of one argument.

joostvanzwieten:
> Any idea why travis is failing? Whatever is going wrong it appears unrelated to the PR. Curious if it persists after a rebuild.

The rebuild works fine.

nutils/solver.py
gertjanvanzwieten: PR #417 seems to be a fairly clean way of doing this?

nutils/solver.py
gertjanvanzwieten: Clear, let's leave it as it is.

gertjanvanzwieten:
> > Any idea why travis is failing? Whatever is going wrong it appears unrelated to the PR. Curious if it persists after a rebuild.
> 
> The rebuild works fine.

Odd. Let's keep an eye on it then to see if the problem resurfaces.

nutils/solver.py
joostvanzwieten: Indeed!

gertjanvanzwieten: Merged.

joostvanzwieten:
I now have `timetarget` as parameter for the name of the `Argument` that represents time, but maybe `timearg` is more suitable? Or simply `time`? Or something else?

PR #426, opened 2019-06-12
Einsum

gertjanvanzwieten:
placeholder solution until a proper function back-end infrastructure is in place

### DIFF
--- a/nutils/function.py
+++ b/nutils/function.py
@@ -55,7 +55,7 @@ def strictevaluable(value):
 def simplified(value):
   return strictevaluable(value).simplified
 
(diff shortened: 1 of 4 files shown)


nutils/function.py
```diff
@@ -2089,8 +2138,8 @@ def __len__(self):
     return 2
 
   def __iter__(self):
-    yield ArrayFromTuple(self, index=0, shape=self.func.shape[:-1], dtype=float)
-    yield ArrayFromTuple(self, index=1, shape=self.func.shape, dtype=float)
+    yield ArrayFromTuple(self, index=0, shape=self.func.shape[:-1], dtype=float if self.symmetric else complex)
+    yield ArrayFromTuple(self, index=1, shape=self.func.shape, dtype=float if self.symmetric else complex)
```
joostvanzwieten: The eigenvectors are also complex if func is complex.

PR #430, opened 2019-06-28
Tensorial

joostvanzwieten:


nutils/expression.py
```diff
@@ -661,11 +661,11 @@ def parse_var(self):
       if '0' <= indices.data <= '9':
         raise _IntermediateError('Expected a non-numeric index, got {!r}.'.format(indices.data), at=indices.pos, count=len(indices.data))
       value = _Array.stack(args, indices.data)
-    elif self._next.type == 'd':
-      self._consume()
```
gertjanvanzwieten: let's keep this in for backwards compatibility

nutils/function.py
```diff
@@ -1079,6 +1037,8 @@ def __init__(self, func:asarray):
   @property
   def simplified(self):
     func = self.func.simplified
+    if func.shape[-2:] == (1,1):
+      return (1 / func).simplified
```
gertjanvanzwieten: suggestion: reciprocal(func).simplified

nutils/function.py
```diff
@@ -1358,6 +1318,8 @@ def __init__(self, func:asarray):
   @property
   def simplified(self):
     func = self.func.simplified
+    if func.shape[-2:] == (1,1):
+      return func[...,0,0]
```
gertjanvanzwieten: return value lacks a .simplified

nutils/sample.py
```diff
@@ -85,16 +85,55 @@ def __init__(self, transforms:types.tuple[transformseq.stricttransforms], points
...
+          detJ *= abs(numpy.dot(J.T, J)[0,0])**0.5
+        else:
+          detJ *= abs(numpy.linalg.det(numpy.dot(J.T, J)))**0.5
+
+        V = numpy.array(V, copy=True, dtype=float)
+        for i in range(root.ndims):
+          if i > 0:
+            V[:,i] -= numpy.dot(V[:,:i], numpy.dot(V[:,:i].T, V[:,i]))
+          V[:,i] /= numpy.linalg.norm(V[:,i])
+        T, N = V[:,:trans[-1].fromdims], V[:,trans[-1].fromdims:]
```
gertjanvanzwieten: shouldn't there be an additional case here for trans[-1].fromdims == root.ndims so that we can take the direct determinant of linear?

nutils/sample.py
```diff
@@ -85,16 +85,55 @@ def __init__(self, transforms:types.tuple[transformseq.stricttransforms], points
...
+        T = numpy.zeros((root.ndims,0))
+        N = numpy.eye(root.ndims)
+      elif not trans:
+        T = numpy.eye(root.ndims)
+        N = numpy.zeros((root.ndims,0))
+      else:
+        V = transform.linearfrom(trans, root.ndims)
+        J = V[:,:trans[-1].fromdims]
+        if J.shape[1] == 1:
+          detJ *= abs(numpy.dot(J.T, J)[0,0])**0.5
```
gertjanvanzwieten: don't you think this 1x1 special case is handled well by linalg.det?

nutils/topology.py
```diff
@@ -878,15 +932,15 @@ class Point(Topology):
...
   def __init__(self, trans, opposite):
     assert trans[-1].fromdims == 0
     references = elementseq.asreferences([element.getsimplex(0)], 0)
-    transforms = transformseq.PlainTransforms((trans,), 0)
-    opposites = transforms if opposite is None else transformseq.PlainTransforms((opposite,), 0)
+    transforms = functools.reduce(transformseq.ProductTransforms, [transformseq.PlainTransforms((transi,), 0) for transi in trans])
+    opposites = functools.reduce(transformseq.ProductTransforms, [transformseq.PlainTransforms((oppi,), 0) for oppi in opposite])
     super().__init__(references, transforms, opposites)
 
 class StructuredLine(Topology):
```
gertjanvanzwieten: Is this currently used? If not, won't it be better to remove it with the aim to transition StructuredTopology into 1D only?

nutils/topology.py
```diff
@@ -2154,7 +2151,7 @@ class RevolutionTopology(Topology):
 
   def __init__(self):
     self._root = transform.Identifier(1, 'angle')
```
gertjanvanzwieten: is this correct?

nutils/topology.py
joostvanzwieten: Removed.

nutils/topology.py
joostvanzwieten: Removed.

joostvanzwieten:
Replaced by #611.

PR #451, opened 2019-07-24
fixed and fallback lengths in expressions

joostvanzwieten:


CHANGELOG.rst
```diff
@@ -8,6 +8,44 @@ inverse chronological order.
...
+
+      >>> ns = nutils.function.Namespace(fixed_lengths=dict(i=2))
+      >>> ns.a = numpy.array([1,2,3])
+      >>> 'a_i' @ ns
+      Traceback (most recent call last):
+      ...
+      ExpressionSyntaxError: Length of index i is fixed at 2 but the expression has length 3.
+      a_i
+        ^
+
```
gertjanvanzwieten: I suggest we keep changelog entries fairly brief and link to the API reference (function.Namespace) for details. In any case the last paragraph (incompatible lengths) should be included in the Namespace docstring.

nutils/function.py
```diff
@@ -4179,6 +4179,29 @@ class Namespace:
...
+  >>> ns_fixed = function.Namespace(fixed_lengths=dict(i=2))
+  >>> ns_fixed.eval_ij('δ_ij')
+  Array<2,2>
+
+  The second is to define a fallback length via the ``fallback_length`` argument:
+
+  >>> ns_fallback = function.Namespace(fallback_length=2)
+  >>> ns_fallback.eval_ij('δ_ij')
+  Array<2,2>
+
```
gertjanvanzwieten: Just a thought: should we support the syntax ``fixed_lengths=dict(ijk=3, pqr=2)`` for ease of use? Since indices are always single character there are no ambiguities. We could unpack this in Namespace, I don't think there is value in passing it on to expression in this form.

CHANGELOG.rst
joostvanzwieten: I've reduced the changelog entry.

nutils/function.py
joostvanzwieten: Done. Even shorter alternative: `Namespace(length_ijk=3, length_pqr=2)`.

PR #452, opened 2019-07-26
Transforms updates

joostvanzwieten:
The goal of this PR is to eliminate `PlainTransforms` in the `UnionTopology` and in the `boundary` and `interfaces` attributes of the `Topology` base class.

### DIFF
--- a/nutils/topology.py
+++ b/nutils/topology.py
@@ -818,7 +818,7 @@ class EmptyTopology(Topology):
 
   @types.apply_annotations
   def __init__(self, ndims:types.strictint):
-    super().__init__(elementseq.PlainReferences((), ndims), transformseq.PlainTransforms((), ndims), transformseq.PlainTransforms((), ndims))
+    super().__init__(elementseq.EmptyReferences(ndims), transformseq.PlainTransforms((), ndims), transformseq.PlainTransforms((), ndims))
 
   def __or__(self, other):
     assert self.ndims == other.ndims
--- a/nutils/topology.py
+++ b/nutils/topology.py
@@ -20,12 +20,11 @@
 
 """
 The topology module defines the topology objects, notably the
-:class:`StructuredTopology` and :class:`UnstructuredTopology`. Maintaining
-strict separation of topological and geometrical information, the topology
-represents a set of elements and their interconnectivity, boundaries,
-refinements, subtopologies etc, but not their positioning in physical space. The
-dimension of the topology represents the dimension of its elements, not that of
-the the space they are embedded in.
+:class:`StructuredTopology`. Maintaining strict separation of topological and
+geometrical information, the topology represents a set of elements and their
+interconnectivity, boundaries, refinements, subtopologies etc, but not their
+positioning in physical space. The dimension of the topology represents the
+dimension of its elements, not that of the the space they are embedded in.
 
 The primary role of topologies is to form a domain for :mod:`nutils.function`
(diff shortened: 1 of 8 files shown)


nutils/topology.py
```diff
@@ -609,38 +608,63 @@ def boundary(self):
     '''
```
gertjanvanzwieten: Should we use the opportunity to move this to ConnectedTopology?

nutils/topology.py
```diff
@@ -609,38 +608,63 @@ def boundary(self):
...
+    if refs_touched:
+      references = elementseq.asreferences(references, self.ndims-1)
+    else:
+      references = self.references.edges[selection]
+    transforms = self.transforms.edges(self.references)[selection]
+    return Topology(references, transforms, transforms)
 
   @property
   @log.withcontext
   def interfaces(self):
```
gertjanvanzwieten: idem boundary

tests/test_elementseq.py
```diff
@@ -82,44 +81,68 @@ def test_mul_references(self):
...
-  check=[square*line]*2+[triangle*line]*2,
-  checkndims=3)
-common(
-  'ChildReferences',
-  seq=nutils.elementseq.ChildReferences(nutils.elementseq.PlainReferences([square, triangle], 2)),
-  check=[square]*4+[triangle]*4,
-  checkndims=2)
+class EmptyReferences(TestCase, Common):
+  @classmethod
+  def setUpClass(cls):
```
gertjanvanzwieten: why not just setUp instead of setUpClass? (applies to all following)

tests/test_transformseq.py
```diff
@@ -153,115 +160,167 @@ def test_refined(self):
...
-common(
-  'StructuredTransforms:2D,refined',
-  seq=nutils.transformseq.StructuredTransforms(x2, [nutils.transformseq.DimAxis(0,2,False),nutils.transformseq.DimAxis(2,4,False)], 1),
-  check=[(x2,s01,c00),(x2,s01,c01),(x2,s01,c10),(x2,s01,c11)],
-  checkmissing=[(x2,s00,c00)],
-  checkrefs=nutils.elementseq.asreferences([square]*4, 2),
-  checkfromdims=2)
+class EmptyTransforms(TestCase, Common, Edges):
+  @classmethod
+  def setUpClass(cls):
```
gertjanvanzwieten: idem test_elementseq

tests/test_elementseq.py
joostvanzwieten: done

tests/test_transformseq.py
joostvanzwieten: done

nutils/topology.py
joostvanzwieten: That requires a bit more work than just moving the properties. The `ConnectedTopology` takes a precomputed connectivity. The `SimplexTopology` does not define a `boundary` or `interfaces` itself, but subclassing from `ConnectedTopology` requires computing the connectivity in `__init__`. We could add a, say, `ConnectedTopologyBase` with an abstract property `connectivity`, move the `boundary` and `interfaces` properties to this class and let `SimplexTopology` and `ConnectedTopology` derive from `ConnectedTopologyBase`.

gertjanvanzwieten: I see, agreed. ConnectedTopologyBase seems to be the right approach but let's leave it for later.

PR #456, opened 2019-10-01
Stringly

gertjanvanzwieten:


nutils/types.py
```diff
@@ -1475,16 +1496,22 @@ def _parse(cls, s):
...
+
+  @classmethod
+  def __stringly_loads__(cls, s):
+    return cls(s)
+
+  @classmethod
+  def __stringly_dumps__(cls, v):
+    if not hasattr(cls, 'unit'):
+      raise Exception('only bound units can be dumped')
+    return '{:f}'.format(v/cls.value).strip('0').rstrip('.') + cls.unit
```
joostvanzwieten: Maybe you should add an `if not isinstance(v, float): raise Exception/stringly.error.SerializationError` here.

PR #457, opened 2019-10-10
Stringly

gertjanvanzwieten:


nutils/types.py
```diff
@@ -1444,7 +1444,7 @@ def __getitem__(self, unit):
     '''
     if unit[0] in '1234567890.*':
       raise ValueError('unit cannot start with a numeral')
-    return type(unit, (float,), dict(__stringly_loads__=classmethod(self._loads), __stringly_dumps__=classmethod(self._dumps)))
+    return type('unit:'+unit, (float,), dict(__unit__=unit, __stringly_loads__=classmethod(self._loads), __stringly_dumps__=classmethod(self._dumps)))
```
joostvanzwieten: Claiming `__unit__` is dangerous as [future versions of Python might use it for a different purpose](https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers). Wouldn't `unit` do?

gertjanvanzwieten: Should do, I'll change it.

PR #476, opened 2019-11-12
move droptol from newton, minimize to optimize

gertjanvanzwieten:


nutils/solver.py
```diff
@@ -583,7 +572,7 @@ def resume_index(self, history, index):
 
 
 @log.withcontext
-def optimize(target:types.strictstr, functional:sample.strictintegral, *, newtontol:types.strictfloat=0., arguments:argdict={}, **kwargs):
+def optimize(target:types.strictstr, functional:sample.strictintegral, *, newtontol:types.strictfloat=0., arguments:argdict={}, droptol=None, constrain=None, solveargs={}, lhs0=None, **kwargs):
```
joostvanzwieten: Since `optimize` doesn't use `newton` anymore, I would add the `function.cache` decorator.

gertjanvanzwieten: fixed

PR #508, opened 2020-02-26
fix missing boundary term in compatible NS examples

gertjanvanzwieten:


bartimore:
I think you confused (the mathematician) "Nitsche" with (the philsopher) "Nietzsche" ;)

gertjanvanzwieten:
@bartimore Good catch, thanks! Fixed now.

PR #516, opened 2020-03-27
switch to meshio for parsing gmsh .msh files

gertjanvanzwieten:


nutils/util.py
```diff
@@ -419,4 +419,30 @@ def readtext(path):
...
+    return path.open('rb')
+
+  if isinstance(path, str):
+    return open(path, 'rb')
+
+  if isinstance(path, io.BufferedIOBase):
+    return contextlib.nullcontext(path) if hasattr(contextlib, 'nullcontext') \
+      else contextlib.contextmanager(iter)([path]) # Python <= 3.6
+
+  raise TypeError('readtext requires a path-like or file-like argument')
```
joostvanzwieten, line 446: The error message should read `binaryfile` instead of `readtext`.

nutils/mesh.py
```diff
@@ -496,31 +338,38 @@ def parsegmsh(mshdata):
...
+  # for linear geometries or larger for higher order geometries. Since meshio
+  # stores nodes by simplex type and cell, simplex types are mapped to
+  # dimensions and gathered, after which cells are concatenated under the
+  # assumption that there is only one simplex type per dimension.
+  nodes = {('ver','lin','tri','tet').index(typename[:3]): numpy.concatenate(datas, axis=0)
+    for typename, datas in util.gather((cells.type, cells.data) for cells in msh.cells)}
+
+  # Identities is a 2d int-aray such that pairs slave nodes, identities[:,1],
+  # to master nodes, identities[:,0]. For the topological connectivity, all
+  # slaves in the nodes arrays will be replaced by there master counterpart.
```
joostvanzwieten, line 362: `there` should be `their`

nutils/types.py
```diff
@@ -280,28 +280,33 @@ def nutils_hash(data):
...
+    h.update(hashlib.sha1(data).digest())
   elif t is builtins.tuple:
-    hargs = map(nutils_hash, data)
+    for item in data:
+      h.update(nutils_hash(item))
   elif t is frozenset:
-    hargs = sorted(map(nutils_hash, data))
+    for item in sorted(map(nutils_hash, data)):
+      h.update(item)
+  elif issubclass(t, io.BufferedIOBase):
```
joostvanzwieten, line 302: Should we also test that `t.writable()` is false?

nutils/util.py
gertjanvanzwieten: fixed

nutils/mesh.py
gertjanvanzwieten: fixed

nutils/types.py
gertjanvanzwieten: fixed; also added seekable

PR #522, opened 2020-04-07
add webhook repo_token

gertjanvanzwieten:
Experiment to see if this makes coveralls.io happy, even though the [documentation](https://docs.coveralls.io/parallel-build-webhook) states that "with public repos, you can omit the repo token".


gertjanvanzwieten:
Experiment failed

PR #527, opened 2020-04-14
disable parallel tests on macOS

joostvanzwieten:
When running all unittests on macOS with matplotlib installed, sometimes a
forked child segfaults with the message

    libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Couldn't close file

This patch sweeps the problem under the carpet by disabling unittests with fork
on macOS.

### DIFF
--- a/tests/test_parallel.py
+++ b/tests/test_parallel.py
@@ -1,8 +1,9 @@
-import unittest, os, multiprocessing, time
+import unittest, os, multiprocessing, time, sys
 from nutils import parallel
 
 canfork = hasattr(os, 'fork')
 
+@unittest.skipIf(sys.platform == 'darwin', 'fork is unreliable (in combination with matplotlib)')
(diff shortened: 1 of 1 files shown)


tests/test_parallel.py
```diff
@@ -1,11 +1,13 @@
...
 from nutils import parallel
 
 canfork = hasattr(os, 'fork')
 
 class Test(unittest.TestCase):
 
   def setUp(self):
+    if sys.platform == 'darwin':
+      self.skipTest('fork is unreliable (in combination with matplotlib)')
     parallel._maxprocs = 3
```
gertjanvanzwieten, line 11: Isn't it cleaner to decorate the class with @skipIf? That should work according to https://docs.python.org/3.5/library/unittest.html#skipping-tests-and-expected-failures

joostvanzwieten: done

PR #529, opened 2020-04-16
add pointsseq module

joostvanzwieten:


nutils/pointsseq.py
```diff
@@ -0,0 +1,648 @@
...
+  '''
+
+  __slots__ = '_items'
+
+  @types.apply_annotations
+  def __init__(self, items: types.tuple[points.strictpoints], ndims: types.strictint):
+    if len(items) == 0:
+      warnings.warn('expected at least 2 items but got 0; use `Empty(ndims)` instead', warnings.NutilsInefficiencyWarning)
+    elif len(items) == 1:
+      warnings.warn('expected at least 2 items but got 1; use `Uniform(items[0], 1)` instead', warnings.NutilsInefficiencyWarning)
```
gertjanvanzwieten, line 293: should we check instead if all elements are equal, since this situation should have been steered to Uniform by aspoints? (same holds in elementsseq)

nutils/pointsseq.py
```diff
@@ -0,0 +1,648 @@
...
+    elif not self:
+      return other
+    elif self == other:
+      return Repeat(self, 2)
+    else:
+      return Chain(self, other).balanced
+
+  @property
+  def isuniform(self):
+    '''``True`` if all points in this sequence are equal.'''
```
gertjanvanzwieten, line 223: since we expect all sequences for which isuniform is True to be Uniform (otherwise they're anyhow inefficient) can't we just do isinstance checks from within Uniform instead of having this property? just a thought; maybe the main reason for having this is the inefficiencywarnings.

nutils/pointsseq.py
```diff
@@ -0,0 +1,648 @@
...
+
+  __slots__ = '_sequence1', '_sequence2'
+  __cache__ = 'tri', 'hull'
+
+  @types.apply_annotations
+  def __init__(self, sequence1: strictpointsseq, sequence2: strictpointsseq):
+    if sequence1.ndims != sequence2.ndims:
+      raise ValueError('cannot chain sequences with different ndims')
+    elif sequence1.isuniform and sequence2.isuniform and sequence1.get(0) == sequence2.get(0):
+      warnings.warn('both sequences are uniform sequences of the same item, use `Uniform(sequence1.get(0), len(sequence1) + len(sequence2))` instead', warnings.NutilsInefficiencyWarning)
```
gertjanvanzwieten, line 548: we can also warn for the situation sequence1 == sequence2: should be a Repeat.

nutils/elementseq.py
```diff
@@ -229,159 +321,108 @@ class UniformReferences(References):
...
 
   Parameters
   ----------
   parent : :class:`References`
-      The transforms to subset.
+      The sequence to select from.
   indices : one-dimensional array of :class:`int`\\s
-      Indices of ``parent`` that form this selection.
+      The indices of ``parent`` references to select.
   '''
```
gertjanvanzwieten, line 392: is it correct that Take does not define getpoints?

gertjanvanzwieten:
Commit fc217529 contains an alternative approach to tree balancing, but I'm not sure if it's an improvement. Both this implementation and the original suffer from potential inefficiency warnings in case a Chain is formed bypassing the special cases in the chain method.

nutils/pointsseq.py
joostvanzwieten: done

nutils/pointsseq.py
joostvanzwieten: With the private implementations the insinstance check is not possible anymore.

nutils/pointsseq.py
joostvanzwieten: done

nutils/elementseq.py
joostvanzwieten: This class is probably only used with `_Plain` as parent, in which case `parent.getpoints(...).take(...)` is less efficient than the base implementation.

PR #542, opened 2020-05-08
Matrix

gertjanvanzwieten:


nutils/matrix/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+    return util.joined_contexts(_assemble.sets(assemble), setthreading(args))
+  elif name == 'auto':
+    try:
+      from ._mkl import assemble
+    except BackendNotAvailable:
+      try:
+        from ._scipy import assemble
+      except BackendNotAvailable:
+        from ._numpy import assemble
+    return _assemble.sets(assemble)
```
joostvanzwieten, line 62: With MKL installed `backend('auto')` does not call `_mkl.setthreading`.

joostvanzwieten: Wouldn't it be simpler to move the backend-specific setters to the backends?

PR #574, opened 2020-08-12
Lower

joostvanzwieten:


nutils/function.py
```diff
@@ -37,7 +37,26 @@
...
+        The dimension of the :class:`~nutils.sample.Sample` on which the
+        resulting :class:`nutils.evaluable.Array` will be evaluated.
+    opposite : :class:`bool`
+        Indicates which transform chain to use when evaluating the resulting
+        :class:`~nutils.evaluable.Array`. This has no effect when there is only
+        one transform chain.
+    with_points : :class:`bool`
+        If true the resulting :class:`~nutils.evaluable.Array` will have a leading points axis.
+    npoints : :class:`int` or :class:`nutils.evaluable.Array`
+        The length of the points axis.
```
gertjanvanzwieten, line 55: can't this be combined with with_points?

nutils/function.py
```diff
@@ -626,7 +663,7 @@ def __init__(self, ndims: int) -> None:
   def __getnewargs__(self):
     return self.shape[0],
 
-  def prepare_eval(self, **kwargs: Any) -> evaluable.Array:
+  def prepare_eval(self, *, with_points: bool = True, **kwargs: Any) -> evaluable.Array:
```
gertjanvanzwieten, line 666: what is the point of adding with_points here?

gertjanvanzwieten (commented):
Some minor things for now; I still need to study the new function.py in detail. Looks great though!

gertjanvanzwieten (approved):

nutils/function.py
```diff
@@ -18,2906 +18,2514 @@
...
-      try:
-        f, args = obj.__reduce__()
-      except: # obj cannot be reduced into a constructor and its arguments
-        rstack.append(obj)
-      else:
-        fstack.append(f)
-        fstack.append(len(args))
-        fstack.append(recreate)
-        fstack.extend(args)
+      raise ValueError
```
gertjanvanzwieten, line 118: Let's add a message here such as ValueError('cannot convert {} to Array'.format(type(__value).__name). Also perhaps we can switch to single underscores for __value so it becomes printable in pdb.

gertjanvanzwieten (changes requested):

nutils/function.py
joostvanzwieten: Done. Mypy treats parameters with double underscores as positional only (see https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#functions , function `quux`).

gertjanvanzwieten: Ah yes, that was it.

gertjanvanzwieten (approved):

PR #580, opened 2020-10-02
Docker

joostvanzwieten:


nutils/matrix/__init__.py
```diff
@@ -89,4 +89,6 @@ def Scipy():
 def MKL(threading:str=None):
   return _helper('MKL', threading)
 
+backend(os.environ.get('NUTILS_MATRIX') or 'auto')
```
gertjanvanzwieten, line 92: Did you test this? Backend returns an unentered context so I don't think this is currently effective. Rather than entering it I would propose to set the _assemble default based on the envvar: `_assemble = util.settable(_import_backend(os.environ.get('NUTILS_MATRIX', 'auto')).assemble)`. The only downside is that we cannot pass the threading argument to mkl, but that might be solved using the MKL_THREADING_LAYER envvar: https://software.intel.com/content/www/us/en/develop/documentation/mkl-macos-developer-guide/top/linking-your-application-with-the-intel-math-kernel-library/linking-in-detail/dynamically-selecting-the-interface-and-threading-layer.html (not tested).

devtools/__init__.py
```diff
@@ -0,0 +1,49 @@
+#! /usr/bin/env python3
+
+import sys, os, subprocess, shlex, typing
+from typing import Mapping, Dict, Optional
+
+GITHUB_ACTIONS = bool(os.environ.get('GITHUB_ACTIONS'))
+OFFICIAL_CONTAINER_REPO = 'ghcr.io/evalf/nutils'
```
gertjanvanzwieten, line 7: perhaps it's cleaner to move this to `container/__init__.py`?

devtools/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+    else:
+      orig_excepthook(exctype, value, tb)
+
+  sys.excepthook = new_excepthook
+
+else:
+
+  from . import _log_default as log
+
+class Error(Exception): pass
```
gertjanvanzwieten, line 27: For simplicity we could consider removing `Error` and use `sys.exit` instead, catching `SystemExit` in `new_excepthook`. This also removes tracebacks in the non-gha scenario for valid failure paths.

devtools/_git.py
```diff
@@ -0,0 +1,36 @@
+from typing import Generator
+from contextlib import contextmanager
+from subprocess import CompletedProcess
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from . import run
+
+class Git:
```
gertjanvanzwieten, line 8: Since `Git` is only used with its default argument we could consider limiting its use to that special case and turning its methods into module methods (similar to the logs) so we don't need to instantiate it - unless we anticipate this situation may change in future.

devtools/_log_gha.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any
+
+def debug(*args: Any) -> None:
+  print(*args)
```
gertjanvanzwieten, line 4: Isn't that simply `debug = print`?

README.md
```diff
@@ -25,54 +25,74 @@ applications.
...
 
-With Python installed, the recommended way to install Nutils is to clone [the
-repository][3], followed by an editable installation using [pip][4] (included
-in the standard installer):
+With Python installed, the recommended way to install the latest stable version
+of Nutils is through [pip][4] (included in the standard installer):
 
-    $ git clone https://github.com/nutils/nutils.git
-    $ python3 -m pip install --user --editable nutils
+    python3 -m pip install --user nutils
```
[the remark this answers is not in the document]
joostvanzwieten: With `pip install user` we pull Nutils from PyPI.

devtools/__init__.py
joostvanzwieten: Done.

devtools/__init__.py
joostvanzwieten: I'll replace `raise Error` with `raise SystemExit` if that's OK.

devtools/_git.py
joostvanzwieten: `Git.worktree` makes a new instance of `Git`.

devtools/_log_gha.py
joostvanzwieten: Yes!

README.md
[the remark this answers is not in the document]
gertjanvanzwieten: Of course - I was still reading this as a (non editable) install from a clone.

devtools/__init__.py
gertjanvanzwieten: sure, I believe there's no diference

devtools/_git.py
gertjanvanzwieten: Check

gertjanvanzwieten (approved):

PR #584, opened 2020-10-14
fix: index dtype sparse data structure

timovanopstal:
Playing around with the number of elements in the tutorial and stumbled across a cryptic `Exception: all blocks should be either 1d-sparse or 0` which can be traced back to selecting a wrong data type for indices of sparse data structures, reproduced with the MWE below. PR is a fix, but needs a check from someone who actually knows what is going on.

```
from nutils import mesh, sample, sparse
domain, geom = mesh.line(255)
basis = domain.basis('std', 1)
integral = domain.integral(basis, degree=1)
# stripped down version of solver._integrate_blocks
data = sample.eval_integrals_sparse(integral)
vector = sparse.block(data) # fails due to chosen index dtype
```

### DIFF
--- a/nutils/sparse.py
+++ b/nutils/sparse.py
@@ -346,7 +346,7 @@ def _dtype(itype, vtype):
   return numpy.dtype([('index', itype), ('value', vtype)])
 
 def _uint(n):
-  return numpy.dtype('>u'+str(1 if n <= 256 else 2 if n <= 256**2 else 4 if n <= 256**4 else 8))
+  return numpy.dtype('>u'+str(1 if n < 256 else 2 if n < 256**2 else 4 if n < 256**4 else 8))
 
 def _resize(data, n):
   if data.base is not None:


gertjanvanzwieten:
Reviewing `_uint` it does look correct to me in its present state: `n` represents an exclusive bound so it should not itself fit in the integer. Running the code I also cannot seem to trigger the issue. Can you elaborate on in what way the last line fails? Note that `sparse.block` (which is a concatenator) in this case simply returns `data[0]` since `data` contains only a single block.

There have been two recent issues relating to integer size: #579 fixed `issparsedtype` which failed to recognize the result of `_uint(256)` as a valid type, and #583 fixed an integer truncation issue related to MKL's 1-based sparse matrix representation. Both are merged in master, but both fairly recent. Could it be that you run into any of these?

timovanopstal:
Seems to have been #579 indeed, my bad for not updating!

gertjanvanzwieten:
No worries. On the contrary, thanks for taking the time to create a PR here (though I'm glad this turned out not to be yet another issue related to uint)

PR #588, opened 2020-10-19
restore function.heaviside function

gertjanvanzwieten:


nutils/function.py
```diff
@@ -2452,6 +2452,27 @@ def partition(f: IntoArray, *levels: float) -> Sequence[Array]:
   steps = map(subtract, signs[:-1], signs[1:])
   return [.5 - .5 * signs[0]] + [.5 * step for step in steps] + [.5 + .5 * signs[-1]]
 
+def heaviside(f: IntoArray):
+  '''
+  Create a heaviside function based on a levelset:
+  `H(f) = { 0: f < 0; 0.5: f == 0; 1: f > 0`.
```
joostvanzwieten, line 2458: Use double backticks. For coherence with the rest of the module, perhaps you can move the first line with text to the same line as the triple quotes and move the mathematical definition to a separate paragraph, maybe even in a `.. math::` block.

nutils/function.py
```diff
@@ -2452,6 +2452,27 @@ def partition(f: IntoArray, *levels: float) -> Sequence[Array]:
...
+
+  Heaviside is a special case of :func:`partition`.
```
joostvanzwieten, line 2460: Maybe you can move this to section `See Also` below `Returns`. Example:

```
See Also
--------

:func:`partition`: generalized version of :func:`heaviside`
:func:`sign`: like :func:`heaviside` but with different levels
```

nutils/function.py
```diff
@@ -2452,6 +2452,27 @@ def partition(f: IntoArray, *levels: float) -> Sequence[Array]:
...
+  f : :class:`Array`
+      Scalar-valued function
+
+  Returns
+  -------
+  :class:`Array`
+      The heaviside function.
+  '''
+
+  left, right = partition(f, 0)
```
joostvanzwieten, line 2473: Since `partition` itself uses `sign`, wouldn't it make sense to use `.5 + .5 * sign(f)` instead?

gertjanvanzwieten: absolutely! :)

nutils/function.py
```diff
@@ -2452,6 +2452,34 @@ def partition(f: IntoArray, *levels: float) -> Sequence[Array]:
   steps = map(subtract, signs[:-1], signs[1:])
   return [.5 - .5 * signs[0]] + [.5 * step for step in steps] + [.5 + .5 * signs[-1]]
 
+def heaviside(f: IntoArray):
+  '''Create a heaviside step-function based on a scalar function f.
+
+  .. math:: H(f) &= 0     && f < 0
+
+            H(f) &= 0.5   && f == 0
```
joostvanzwieten, line 2460: In math mode `==` seems non-standard. I suggest to replace `==` with `=`.

gertjanvanzwieten: done

joostvanzwieten (approved):

PR #593, opened 2020-12-10
remove locate args ischeme and scale, add maxdist

gertjanvanzwieten:
Formerly locate made an initial selection of candidate elements based on
(scaled) bounding boxes. The new implementation simply orders elements by
centroid proximity, with an option to set a cutoff radius in case fast failure
is required.

### DIFF
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -1,7 +1,7 @@
 from nutils import *
 from nutils.testing import *
 from nutils.elementseq import References
-import numpy, copy, sys, pickle, subprocess, base64, itertools, os
+import numpy, copy, sys, pickle, subprocess, base64, itertools, os, unittest
 
 class TopologyAssertions:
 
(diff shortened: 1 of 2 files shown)


tests/test_topology.py
```diff
@@ -352,6 +352,16 @@ def test(self):
     located = sample.eval(self.geom)
     self.assertAllAlmostEqual(located, target)
 
+  def test_maxdist(self):
+    if self.etype == 'square' and self.mode != 'nonlinear':
+      raise unittest.SkipTest('maxdist does not affect rectilinear meshes')
```
joostvanzwieten, line 357: With `unittest.SkipTest` this test will be marked as skipped, which we usually only do when dependencies are missing. Maybe you can replace this with

```python
@testing.enable_if(lambda etype, mode, **kwargs: etype != 'square' or mode != 'nonlinear')
def test_maxdist(self):
  target = numpy.array([(.2,.3), (.1,.9), (0,1)])
  ...
```

joostvanzwieten (changes requested):

joostvanzwieten (approved):

PR #603, opened 2020-12-15
Fix loop concatenate sparse axes

joostvanzwieten:


nutils/sample.py
```diff
@@ -125,6 +125,9 @@ def getindex(self, ielem):
 
     raise NotImplementedError
 
+  def _get_evaluable_indices(self, ielem):
+    return evaluable.ElemwiseFromCallable(self.getindex, ielem, self.points.get_evaluable_coords(ielem).shape[:1], int)
```
gertjanvanzwieten, line 129: Can't `_CustomIndex` implement this as `Elemwise(self._index, ielem, int)`?

joostvanzwieten: In theory: yes. In practice, however, the shape of `Elemwise` doesn't match the shape of the function that will be inflated in `_AtSample`, because the latter has a shape that originates from the `PointsSequence`.

joostvanzwieten:
Update: I've made `Sample.get_evaluable_indices` public, similar to the evaluable variants of coords and weights in the `PointsSequence`.

nutils/sample.py
gertjanvanzwieten: ok

gertjanvanzwieten (approved):

PR #606, opened 2020-12-18
fix __repr__ for frozenarray

soraros:


gertjanvanzwieten (changes requested):
Perhaps you can add a unittest to the frozenarray TestCase in tests/test_types.py for coverage.

PR #607, opened 2020-12-18
Attach dtype to nodes

soraros:


nutils/evaluable.py
```diff
@@ -297,7 +297,7 @@ def _node(self, cache, subgraph, times):
 
   @property
   def _node_details(self):
-    return ''
+    return self.dtype.__name__ if hasattr(self, 'dtype') else ''
```
gertjanvanzwieten, line 300: Rather than putting this in the base class and testing for an attribute I'd say proper way is to place this in `Array._node_details`, or better yet `Array._node` which is where the shape is handled as well. This way operations such as `Transpose` and `Einsum` can add details without losing the type.

tests/test_evaluable.py
```diff
@@ -681,16 +681,16 @@ class asciitree(TestCase):
   def test_asciitree(self):
     f = evaluable.Sin(evaluable.Inflate(1, evaluable.Zeros((), int), 2)**evaluable.Diagonalize(evaluable.Argument('arg', (2,))))
     self.assertEqual(f.asciitree(richoutput=True),
-                     '%0 = Sin; a2,a2\n'
-                     '└ %1 = Power; a2,a2\n'
+                     '%0 = Sin; float; a2,a2\n'
+                     '└ %1 = Power; float; a2,a2\n'
                      '  ├ %2 = Transpose; 1,0; i2,s2\n'
```
gertjanvanzwieten, line 686: (here you lost the dtype because of using `_node_details`)

gertjanvanzwieten (changes requested):

tests/test_evaluable.py
```diff
@@ -722,19 +723,19 @@ def test_loop_concatenate(self):
...
+                     '│ │ └ %A3 = InsertAxis; i:i2\n'
                      '│ │   ├ 1\n'
                      '│ │   └ 2\n'
                      '│ └ %B4 = LoopIndex\n'
                      '│   └ length = 2\n'
-                     '├ stop = %B5 = Take; int\n'
+                     '├ stop = %B5 = Take; i\n'
                      '│ ├ %A2\n'
-                     '│ └ %B6 = Add; int\n'
+                     '│ └ %B6 = Add; i\n'
```
soraros, line 735: @gertjanvanzwieten Is the short form desirable when `shape` is not present?

tests/test_evaluable.py
```diff
@@ -684,18 +684,19 @@ class asciitree(TestCase):
   @unittest.skipIf(sys.version_info < (3, 6), 'test requires dicts maintaining insertion order')
   def test_asciitree(self):
     f = evaluable.Sin(evaluable.Inflate(1, evaluable.Zeros((), int), 2)**evaluable.Diagonalize(evaluable.Argument('arg', (2,))))
+    self.maxDiff = None
```
soraros, line 687: @gertjanvanzwieten What comment should I put here?

gertjanvanzwieten: something to keep us from accidentally deleting it, like `# keep assertEqual from shortening its arguments`

nutils/evaluable.py
```diff
@@ -858,7 +858,7 @@ def __iter__(self):
   __pow__ = power
   __abs__ = lambda self: abs(self)
   __mod__  = lambda self, other: mod(self, other)
-  __str__ = __repr__ = lambda self: '{}.{}<{}>'.format(type(self).__module__, type(self).__name__, ','.join(map(str, self.shape)) if hasattr(self, 'shape') else '?')
+  __str__ = __repr__ = lambda self: '{}.{}<{}{}>'.format(type(self).__module__, type(self).__name__, self.dtype.__name__[0] +':' if hasattr(self, 'dtype') else '', ','.join(map(str, self.shape)) if hasattr(self, 'shape') else '?')
```
gertjanvanzwieten, line 861: I think I would put the `:` in the format string and put a question mark in the uninitialized state same as for the shape. The fact that both the current and proposed implementation keep the colon visible for scalars (e.g. `Sin<f:>`) is not a bad thing in my opinion, in fact I would propose to use precisely the same for the nodes.

nutils/evaluable.py
```diff
@@ -923,7 +923,7 @@ def _node(self, cache, subgraph, times):
     if self in cache:
       return cache[self]
     args = tuple(arg._node(cache, subgraph, times) for arg in self._Evaluable__args)
-    label = '\n'.join(filter(None, (type(self).__name__, self._node_details, ','.join(map(repr, self._axes)))))
+    label = '\n'.join(filter(None, (type(self).__name__, self._node_details, ':'.join(filter(None, (self.dtype.__name__[0], ','.join(map(repr, self._axes))))))))
```
gertjanvanzwieten, line 926: So this is where I personally would just keep the `:` regardless of dimension. In fact - and I think @joostvanzwieten should weight in here - I actually fail to see why we don't just use `label = str(self)` in the `Evaluable._node` base implementation so we don't have to redefine it in `Array`, and automatically get consistency.

tests/test_evaluable.py
gertjanvanzwieten: I would say so - see comments above.

gertjanvanzwieten (approved):

PR #613, opened 2021-01-11
Check evalf

soraros:


.github/workflows/test.yaml
```diff
@@ -168,6 +170,7 @@ jobs:
       - name: Run unit tests
         env:
           _image: ${{ steps.build.outputs.id }}
+          NUTILS_DEBUG: all
```
joostvanzwieten, line 173: I don't think this works. You can set environment variables inside the container with `podman run ... --env ...`:

```sh
podman run --pull=never --rm -v "$PWD/tests:/app/tests:ro" -v "$PWD/examples:/app/examples:ro" --env NUTILS_DEBUG=all "$_image" -m unittest -bq
```

nutils/evaluable.py
```diff
@@ -757,15 +758,34 @@ def _assparse(self):
...
+    def __get__(self, instance, owner):
+      evalf = self.evalf_obj(instance, owner)
+      @functools.wraps(evalf)
+      def evalf_with_check(*args, **kwargs):
+        res = evalf(*args, **kwargs)
+        if hasattr(instance, 'dtype'):
+          assert asdtype(res.dtype) == instance.dtype, ((instance.dtype, res.dtype), instance, res)
+          if hasattr(instance, 'ndim'):
+            assert res.ndim == instance.ndim
+          assert all(m == n for m, n in zip(res.shape, instance.shape) if isinstance(n, int)), 'shape mismatch'
```
joostvanzwieten, line 780: If there is no `ndim`, then there should not be a `shape` either. Is the `if hasattr(instance, 'ndim')` test necessary?

I think not having a `dtype` doesn't imply there isn't a shape. Maybe we should place the `ndim` and `shape` tests under a separate `if hasattr(instance, 'shape')`?

joostvanzwieten: What do you think of the following:

```python
assert not hasattr(instance, 'dtype') or asdtype(res.dtype) == instance.dtype, ((instance.dtype, res.dtype), instance, res)
assert not hasattr(instance, 'ndim') or res.ndim == instance.ndim
assert not hasattr(instance, 'shape') or all(m == n for m, n in zip(res.shape, instance.shape) if isinstance(n, int)), 'shape mismatch'
```

joostvanzwieten:
Can you reorder the commits such that all of them pass the unittests? Fixes first, then adding the checker.

nutils/debug_flags.py
```diff
@@ -0,0 +1,37 @@
...
+_flags = _all if _env == 'all' \
+    else set(filter(None, _env.lower().split(':'))) | _default if __debug__ \
+    else set(filter(None, _env.lower().split(':')))
+assert _flags <= _all, 'Unknown debug flags: {}'.format(_flags - _all)
+
+for f in _all:
+  locals()[f] = f in _flags
+
+if __name__ == '__main__':
+  print('Debug flags set:', ', '.join(_flags))
```
gertjanvanzwieten, line 37: The following code is a bit easier on the eyes:

```
import os, warnings

_flags = dict.fromkeys(os.getenv('NUTILS_DEBUG', '').lower().split(':'), True)
_all = _flags.pop('all', False)

sparse = _flags.pop('sparse', _all  or __debug__) # check sparse chunks in evaluable
lower = _flags.pop('lower', _all or __debug__) # check lowered shape, dtype in function
evalf = _flags.pop('evalf', _all) # check evaluated arrays in evaluable

if _flags:
  warnings.warn('unused debug flags: {}'.format(', '.join(_flags)))
```

soraros: Done, thanks!

.github/workflows/test.yaml
soraros: Done. Thanks!

nutils/evaluable.py
soraros: Done. Thanks!

joostvanzwieten (approved):

nutils/debug_flags.py
```diff
@@ -0,0 +1,31 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import os, warnings
+
+_env = dict.fromkeys(os.getenv('NUTILS_DEBUG', '').lower().split(':'), True)
```
gertjanvanzwieten, line 23: Just noticed we need to filter empty strings; currently the warning is always triggered.
```
_env = dict.fromkeys(filter(None, os.getenv('NUTILS_DEBUG', '').lower().split(':')), True)
```

gertjanvanzwieten (changes requested):
Could you add the filter and rebase? Then I'll merge it as soon as the tests pass again.

nutils/debug_flags.py
soraros: Done. Thanks!

gertjanvanzwieten (approved):

PR #615, opened 2021-01-13
Basis

gertjanvanzwieten:


nutils/function.py
```diff
@@ -2664,10 +2664,10 @@ def get_support(self, dof: Union[numbers.Integral, numpy.ndarray]) -> numpy.ndar
       dof = numpy.unique(dof)
       if dof[0] < 0 or dof[-1] >= self.ndofs:
         raise IndexError('dof out of bounds')
-      if self.get_support == __class__.get_support.__get__(self, __class__):
-        return numpy.unique([ielem for ielem in range(self.nelems) if numpy.in1d(self.get_dofs(ielem), dof, assume_unique=True).any()])
-      else:
-        return numpy.unique(numpy.fromiter(itertools.chain.from_iterable(map(self.get_support, dof)), dtype=int))
```
gertjanvanzwieten, line 2673: what was the reason for this distinction?

nutils/evaluable.py
```diff
@@ -2359,28 +2363,25 @@ def _simplified(self):
...
+    return self.expr
 
-  def _node(self, cache, subgraph, times):
-    if self in cache:
-      return cache[self]
-    elif hasattr(self.arrays, '_node_tuple'):
-      cache[self] = node = self.arrays._node_tuple(cache, subgraph, times)[self.index]
-      return node
-    else:
-      return super()._node(cache, subgraph, times)
```
gertjanvanzwieten, line 2386: Note sure what this does, either, but we might want to restore it.

nutils/numeric.py
```diff
@@ -547,6 +547,33 @@ def asboolean(array, size, ordered=True):
     barray[array] = True
   return barray
 
+def invmap(indices, length, nan=-1):
```
joostvanzwieten, line 550: What about `missing` instead of `nan`?

nutils/numeric.py
```diff
@@ -547,6 +547,33 @@ def asboolean(array, size, ordered=True):
...
+  '''Create inverse index array.
+
+  >>> m = invmap([3,1], length=5)
```
joostvanzwieten, line 553: Suggestion:

```rst
Create inverse index array.

Create the index array ``inverse`` with the given ``length`` such that ``inverse[indices[i]]`` is ``i`` and ``inverse[j]`` is ``nan`` for all ``j`` not in ``indices``. It is an error pass an ``indices`` array with repeated indices, in which case the result is undefined.
```

nutils/numeric.py
```diff
@@ -547,6 +547,33 @@ def asboolean(array, size, ordered=True):
...
+      Target array length; must be larger than max(indices).
+  nan : :class:`int` (default: -1)
+      Value to insert for missing indices.
+
+  Returns
+  -------
+  :class:`numpy.ndarray`
+  '''
+
+  invmap = numpy.repeat(nan, length)
```
joostvanzwieten, line 573: Suggestion: `invmap = numpy.full(length, nan)`.

nutils/topology.py
```diff
@@ -1479,10 +1479,9 @@ def __or__(self, other):
 
   @property
   def connectivity(self):
-    mask = numpy.array([bool(ref) for ref in self.refs] + [False]) # trailing false serves to map -1 to -1
-    renumber = numpy.cumsum(mask)-1
-    renumber[~mask] = -1
-    return tuple(types.frozenarray(renumber.take(ioppelems).tolist() + [-1] * (ref.nedges - len(ioppelems))) for ref, ioppelems in zip(self.refs, self.basetopo.connectivity) if ref)
+    renumber = numeric.invmap([i for i, ref in enumerate(self.refs) if ref], length=len(self.refs)+1, nan=-1) # length=..+1 serves to map -1 to -1
```
joostvanzwieten, line 1482: I'm not sure if it is faster or more readable, but you can replace `[i for i, ref in enumerate(self.refs) if ref]` with `numpy.formiter(self.refs, dtype=bool).nonzero()[0]`.

nutils/function.py
```diff
@@ -2629,11 +2629,11 @@ def lower(self, **kwargs: Any) -> evaluable.Array:
 
   @util.cached_property
   def _computed_support(self) -> Tuple[types.frozenarray, ...]:
-    support = [set() for i in range(self.ndofs)] # type: List[Set[int]]
+    support = [[] for i in range(self.ndofs)] # type: List[List[int]]
     for ielem in range(self.nelems):
       for dof in self.get_dofs(ielem):
-        support[dof].add(ielem)
-    return tuple(types.frozenarray(numpy.fromiter(sorted(ielems), dtype=int), copy=False) for ielems in support)
```
joostvanzwieten, line 2639: I believe the new implementation does not guarantee uniqueness because `self.get_dofs` is not unique under rare conditions:

```python
from nutils import mesh
topo, geom = mesh.rectilinear([2], periodic=[0])
basis = topo.basis('spline', degree=2)
basis.get_dofs(0) # [0, 1, 0]
```

We use `_computed_support` as a fallback in `get_support`, which does guarantee uniqueness. Either we should lift this restriction or make sure `_computed_support` returns unique arrays. I'm in favor of the latter.

nutils/function.py
joostvanzwieten: The if branch is taken if there's no reimplementation of `getsupport`. I believe the reason was to minimize using the fallback `_computed_support`, but in retrospect I don't think it makes sense.

nutils/evaluable.py
```diff
@@ -3668,6 +3668,11 @@ def exp(x):
 def ln(x):
   return Log(x)
 
+def divmod(x, y):
+  div = FloorDivide(*_numpy_align(x, y))
+  mod = x - div * y
```
joostvanzwieten, line 3666: If the reason of not using function `mod` or class `Mod` defined below is to have some sort of canonical representation, shall we then also reimplement `def mod(arg1, arg2):` using `divmod(arg1, arg2)[1]`?

nutils/evaluable.py
```diff
@@ -3685,6 +3685,12 @@ def divmod(x, y):
 def mod(arg1, arg2):
   return Mod(*_numpy_align(arg1, arg2))
 
+def greater(arg1, arg2):
+  return Greater(*_numpy_align(arg1, arg2))
```
joostvanzwieten, line 3673: The reason for existence of `greater` and `less` seems to be `_numpy_align`. Isn't this something we'd like to get rid of?

nutils/function.py
```diff
@@ -2621,11 +2621,17 @@ def __init__(self, ndofs: int, nelems: int, index: Array, coords: Array) -> None
     self.coords = coords
     super().__init__((ndofs,), float)
 
+    self.arg_dofs, self.arg_coeffs = [f.optimized_for_numpy
+      for f in self.f_dofs_coeffs(evaluable.Argument('_index', shape=(), dtype=int))]
```
joostvanzwieten, line 2574: What do you think of making `f_dofs_coeffs` an `evaluable.Tuple`? Then we can write

```python
self.arg_dofs, self.arg_coeffs = self.f_dofs_coeffs(evaluable.Argument('_index', shape=(), dtype=int)).optimized_for_numpy
```

Suggestion for readability: Introduce a variable for the `Argument('_index', ...)`.

Suggestion: Make `self.arg_dofs` and `self.arg_coeffs` private.

joostvanzwieten: Well, that adds a lot of cruft to other places, so forget about the `evaluable.Tuple`. But maybe with a variable for `Argument` we can get rid of the newline in the list comprehension?

nutils/function.py
```diff
@@ -2716,9 +2725,12 @@ def get_dofs(self, ielem: Union[int, numpy.ndarray]) -> numpy.ndarray:
   def get_ndofs(self, ielem: int) -> int:
     '''Return the number of basis functions with support on element ``ielem``.'''
 
-    return len(self.get_dofs(ielem))
+    ielem = numeric.normdim(self.nelems, ielem) # even if ielem goes unused, we still want the bounds check
+    ndofs, = self.arg_dofs.shape
+    if not isinstance(ndofs, int):
+      ndofs = int(ndofs.eval(_index=ielem))
+    return ndofs
```
joostvanzwieten, line 2681: Did you consider something like

```python
def __init__(self, ...):
  ...
  self.arg_ndofs = evaluable.asarray(self.arg_dofs.shape[0])
  ...

def get_ndofs(self, ielem: int) -> int:
  ielem = numeric.normdim(self.nelems, ielem)
  return int(self.arg_ndofs(ielem))
```

nutils/numeric.py
gertjanvanzwieten: done.

nutils/numeric.py
gertjanvanzwieten: done.

nutils/topology.py
gertjanvanzwieten: I prefer the current for readability, and would expect it to be faster as well since the same work is involved in forming the boolean.

nutils/function.py
gertjanvanzwieten: done.

nutils/evaluable.py
gertjanvanzwieten: I'm not sure.. removing `evaluable.Mod` means repeating this logic in function. Maybe it's better to use `Mod` after all, or keep things local to `divmod` as proposed, from the reasoning that divmod likely flags a situation in which div and mod are going to be evaluated together.

nutils/evaluable.py
gertjanvanzwieten: removed.

nutils/function.py
gertjanvanzwieten: done.

nutils/function.py
gertjanvanzwieten: done.

nutils/numeric.py
gertjanvanzwieten: done.

nutils/sample.py
```diff
@@ -336,9 +335,9 @@ def hull(self):
     return self.points.hull
 
   def get_evaluable_indices(self, ielem):
-    npoints = self.points.get_evaluable_coords(ielem).shape[0]
-    offsets = evaluable._SizesToOffsets(evaluable.loop_concatenate(evaluable.InsertAxis(npoints, 1), ielem, self.nelems))
```
joostvanzwieten, line 340: I'm in favor of keeping `_SizesToOffsets` here and removing `offsets`. With `getindex` using `get_evaluable_indices`, there's no reason to keep `offsets` around, which makes the entire `_DefaultIndex` cache-free. I'll admit that the offsets array is small compared to all other arrays we typically encounter during a simulation, but in some academic cases where for example the element matrix turns out to be constant, this might as well be one of the larger arrays.

`_SizesToOffsets`simplifies to a constant if the argument does not vary w.r.t. the loop index. The same happens with `Constant` of course, but that simplification is rather expensive compared with the former.

joostvanzwieten: And computing `_DefaultIndex.offsets` is done by looping over all elements. With `_SizesToOffsets` this loop is completely eliminated if the shapes do not vary.

nutils/evaluable.py
```diff
@@ -75,16 +75,51 @@ def asarray(arg):
...
+    return True
+  if n.arguments != m.arguments:
+    return False
+  if n.isconstant: # implies m.isconstant
+    return int(n) == int(m)
+
+asshape = types.tuple[asindex]
+
+@types.apply_annotations
+def equalshape(N:asshape, M:asshape):
```
soraros, line 107: Can this or `equalindex` be used in evalf checker?

nutils/evaluable.py
```diff
@@ -2327,45 +2358,40 @@ def evalf(self, points, expect):
...
+    unique, indices = util.unique(data)
+  if len(unique) == 1:
+    return Constant(unique[0])
+  shapes = numpy.empty([len(data), ndim], dtype=int)
+  for i, d in enumerate(data):
+    shapes[i] = d.shape
+  shape = [Take(s, index).simplified for s in shapes.T] # use original index to avoid potential inconsistencies with other arrays
+  if len(unique) < len(data):
+    index = Take(types.frozenarray(indices, copy=False), index)
+  concat = types.frozenarray(numpy.concatenate([d.ravel() for d in unique]), dtype=dtype, copy=False)
```
soraros, line 2376:
```suggestion
  concat = types.frozenarray(numpy.concatenate(unique, axis=None), dtype=dtype, copy=False)
```

nutils/sample.py
```diff
@@ -235,22 +234,14 @@ def __rmatmul__(self, func: function.Array) -> function.Array:
...
     '''Basis-like function that for every point in the sample evaluates to the
     unit vector corresponding to its index.'''
 
     index = function.transforms_index(self.transforms[0])
     coords = function.transforms_coords(self.transforms[0], self.ndims)
-    I = function.Elemwise(self.index, index, dtype=int)
-    B = function.Sampled(coords, expect=function.take(self.allcoords, I, axis=0))
-    return function.inflate(B, indices=I, length=self.npoints, axis=0)
+    sampled = function.Sampled(coords, expect=_Coords(self.points, index))
+    return function.inflate(sampled, indices=_Indices(self, index), length=self.npoints, axis=0)
```
joostvanzwieten, line 247: Instead of using `function.inflate`, you can make a single `function.Array` subclass that lowers to an `evaluable.Inflate`.

joostvanzwieten (changes requested):

nutils/evaluable.py
```diff
@@ -2326,24 +2326,30 @@ def evalf(self, points, expect):
...
+    shape = [Take(s, index).simplified for s in shapes.T] # use original index to avoid potential inconsistencies with other arrays
+  if len(unique) < len(data):
+    index = Take(indices, index)
+  concat = numpy.concatenate([numpy.ravel(d) for d in unique])
+  if ndim == 0:
+    return Take(concat, index)
+  cumprod = list(shape)
+  for i in reversed(range(ndim-1)):
+    cumprod[i] *= cumprod[i+1] # work backwards so that the shape check matches in Unravel
+  offsets = numpy.cumsum([0] + [numpy.prod(d.shape, dtype=int) for d in unique[:-1]])
```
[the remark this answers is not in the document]
gertjanvanzwieten: D is a `types.arraydata` which only defines shape and ndim. Maybe we should condider adding size.

nutils/evaluable.py
```diff
@@ -863,6 +863,15 @@ def __iter__(self):
       raise TypeError('iteration over a 0-d array')
     return (self[i,...] for i in range(self.shape[0]))
 
+  def __index__(self):
+    try:
+      index = self.__index
+    except AttributeError:
+      if self.ndim or self.dtype not in (int, bool) or not self.isconstant:
+        raise TypeError('cannot convert {!r} to int'.format(self))
+      index = self.__index = int(self.simplified.eval())
```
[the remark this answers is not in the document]
gertjanvanzwieten: Unfortunately numpy breaks things again: `numpy.array(True).__index__() -> TypeError`

nutils/evaluable.py
```diff
@@ -2326,24 +2326,30 @@ def evalf(self, points, expect):
...
+  if ndim == 0:
+    return Take(concat, index)
+  cumprod = list(shape)
+  for i in reversed(range(ndim-1)):
+    cumprod[i] *= cumprod[i+1] # work backwards so that the shape check matches in Unravel
+  offsets = numpy.cumsum([0] + [numpy.prod(d.shape, dtype=int) for d in unique[:-1]])
+  elemwise = Take(concat, Range(cumprod[0], Take(offsets, index)))
+  for i in range(ndim-1):
+    elemwise = Unravel(elemwise, shape[i], cumprod[i+1])
+  return elemwise
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/evaluable.py
```diff
@@ -2326,24 +2326,30 @@ def evalf(self, points, expect):
...
-
-  def evalf(self, index):
-    return numpy.asarray(self.data[index])
-
-  def _simplified(self):
-    d0 = self.data[0]
-    if all(di == d0 for di in self.data[1:]):
-      return Constant(d0)
+@types.apply_annotations
+def Elemwise(data:types.tuple[types.arraydata], index:asarray, dtype:asdtype, shape:asshape=None):
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/sample.py
```diff
@@ -145,6 +145,14 @@ def get_evaluable_indices(self, ielem):
 
     return evaluable.ElemwiseFromCallable(self.getindex, ielem, self.points.get_evaluable_coords(ielem).shape[:1], int)
 
+  def lower_for_loop(self, func, **kwargs):
```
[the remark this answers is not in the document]
gertjanvanzwieten: I made the method private - is there a reason to prefer module methods?

nutils/evaluable.py
```diff
@@ -3438,26 +3438,14 @@ def _simplified(self):
...
-    if index.dtype != int or index.ndim != 0:
-      raise ValueError('expected an index with dtype int and dimension zero but got {}'.format(index))
-    if not func.ndim:
-      raise ValueError('expected an array with at least one axis')
-    if any(index in n.arguments for n in func.shape[:-1]):
-      raise ValueError('the shape of the function must not depend on the index')
-    if index in length.arguments:
-      raise ValueError('the length of the loop must not depend on the index')
-    if index in cc_length.arguments:
-      raise ValueError('the length of the concatenation axis must not depend on the index')
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/sample.py
gertjanvanzwieten: done

nutils/sample.py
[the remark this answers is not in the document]
joostvanzwieten: Mypy is satisfied with a private method.

joostvanzwieten (approved):

PR #617, opened 2021-01-15
pytest

soraros:
Switch to `pytest` which prints better error message and supports running tests in parallel. With this PR, CI run twice as fast (except for python 3.5, which somehow doesn't work).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,28 +33,28 @@ jobs:
           if-no-files-found: error
   test:
     needs: build-python-package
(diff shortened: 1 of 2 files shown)


joostvanzwieten:
Based on the coverage report it seems that not all tests are run. I'm a bit hesitant to switch to a different test platform for this reason.

gertjanvanzwieten:
I agree with the fix in test_docs so people can use whichever unit testing framework they are comfortable with (I actually have the same in #615) but I'm also not sure about making this the default. As an extra dependency it adds to potential problems and the waiting times are never much of an issue once the code reaches github - though it's news to me that we apparently have two processors available there.

soraros:
Ok. I guess the verdict is that it's not worth it at least for now. I shall close this PR.

PR #619, opened 2021-02-04
Nofrozenarray

gertjanvanzwieten:
Replace `types.frozenarray` (which adds quite a bit of overhead) by a hasheable array wrapper `types.arraydata` for the construction of immutable objects, and a new cache construction `types.lru_cache` that caches return values based on the buffer (i.e. memory address) of input arguments.


nutils/sample.py
```diff
@@ -69,7 +69,7 @@ class Sample(types.Singleton):
 
   @staticmethod
   @types.apply_annotations
-  def new(transforms:types.tuple[transformseq.stricttransforms], points:types.strict[PointsSequence], index:types.tuple[types.frozenarray[int]]=None):
+  def new(transforms:types.tuple[transformseq.stricttransforms], points:types.strict[PointsSequence], index:types.tuple[numeric.wrapped]=None):
```
soraros, line 72: A `dtype` check is removed here. Should the behaviour of `_frozenarraymeta[dtype]` be reintroduced?

gertjanvanzwieten: That check was supposed to go to the constructor, indeed, good catch.

nutils/numeric.py
```diff
@@ -570,4 +624,70 @@ def levicivita(n: int, dtype=float):
...
+  >>> w = wrapped(a)
+  >>> w == wrapped([1,2,4]) # we cannot do this on a
+  False
+  >>> w.asarray
+  array([1, 2, 3])
+  '''
+
+  __slots__ = 'dtype', 'shape', 'bytes'
+
+  def __init__(self, array):
```
soraros, line 653: Is it better to implement a `__new__` instead so that for `b = wrapped(np.array(...))` we have `wrapped(b) is b`.

gertjanvanzwieten: In this case `__new__` is a bit messy in combination with pickle as that calls `__new__` without arguments followed by `__setstate__` (unless we use `__reduce__`, but then we need to make sure that we do not mistake the state tuple for array contents). Deduplication does matter, though - see the new command in `__setstate__`, I solved it now with a separate function `aswrapped` which I think is the most straightforward way out.

gertjanvanzwieten: The latest implementation of `numeric.wrapped` builds on `types.Immutable`.

nutils/numeric.py
```diff
@@ -23,7 +23,7 @@
 """
 
 from . import types, warnings
-import numpy, numbers, builtins, collections.abc
+import numpy, numbers, builtins, collections.abc, hashlib
```
soraros: This is not used.

nutils/numeric.py
```diff
@@ -24,6 +24,70 @@
...
+    '''Replace the obj attribute (dangerous!)
+
+    The buffer's ``obj`` attribute serves to keep the object alive that owns
+    the memory block. This method may be used to swap the object by another
+    that serves the same purpose. Note that if the new object fails to keep the
+    memory alive this may lead to fatal errors.
+    '''
+
+    pythonapi.Py_DecRef(py_object(self.obj))
+    pythonapi.Py_IncRef(py_object(obj))
```
soraros, line 89: Should we increase the reference count before decrease in case `obj` and `self.obj` refer to the same thing?

gertjanvanzwieten: We could, but since the function frame has its own reference it should be impossible that `obj` gets released between these two lines.

nutils/numeric.py
gertjanvanzwieten: Update: added a commit to have `types.Immutable` support `__new__`, removing the need for `aswrapped`.

nutils/numeric.py
```diff
@@ -439,6 +439,61 @@ def binom(n, k):
...
 
+def full(shape, fill_value):
+  return numpy.lib.stride_tricks.as_strided(fill_value, shape, [0]*len(shape), writeable=False)
+
+def writeable(array):
+  '''Test if any base is writeable.'''
+
+  return array.flags.writeable or isinstance(array.base, numpy.ndarray) and writeable(array.base)
+
+def freeze(array):
```
soraros, line 450: Should this also deal with views created by `stride_tricks` as in `types._deepfreeze`? And if so should this function be called there?

gertjanvanzwieten: Yeah I'm struggling a bit with choosing the right modules, but my intention is indeed to keep only deepfreeze. I'm now thinking of moving `Py_buffer` to util and `keepfrozen` to types.

nutils/numeric.py
```diff
@@ -439,6 +439,61 @@ def binom(n, k):
...
+    except KeyError:
+      v = cache[key] = func(*args)
+      freeze(v)
+      pop = functools.partial(cache.pop, key)
+      for buf in key:
+        if isinstance(buf, Py_buffer):
+          # replace the buffer exporter by a weak reference of the lowest base,
+          # with a callback at destruction to remove the cache entry
+          obj = buf.obj
+          while isinstance(obj, numpy.ndarray) and obj.base is not None:
```
soraros, line 490: Should it be `while isinstance(obj, (numpy.ndarray, numpy.lib.stride_tricks.DummyArray)) ...` as now cache for views created by `stride_tricks` will be poped earlier?

nutils/types.py
```diff
@@ -1481,6 +1481,46 @@ def __setitem__(self, item, value):
...
+      for buf in key:
+        if isinstance(buf, Py_buffer):
+          # replace the buffer exporter by a weak reference of the lowest base,
+          # with a callback at destruction to remove the cache entry
+          obj = buf.obj
+          while isinstance(obj, numpy.ndarray) and obj.base is not None:
+            obj = obj.base
+          buf.replace_obj(weakref.ref(obj, pop))
+    return v
+
```
soraros, line 1308: We could make `cache` an attribute of `wrapper` to enable better debugging.

nutils/numeric.py
soraros: I think you missed this one in the latest version.

gertjanvanzwieten: Indeed! Thanks.

nutils/types.py
gertjanvanzwieten: I experimented with making `lru_cache` a class but it added noticeable overhead, probably due to the logic required for correctly handling bound methods. Instead I'm thinking in the direction of adding caching dictionaries to a global registery (including those of `typing.CacheMeta`) so we have a single place to search for memory leaks.

soraros: What I meant is actually just put `wrapper._cache = cache` before return so cache is exposed and always link to the cached function.
Regarding a single point for debugging, I guess it's easy enough to also attach a UUID to the function(is a function object hashable?) and use that as key and attach them to a global registry: `{uuid: (f, f._cache)}` or simply `{uuid: f}`.

gertjanvanzwieten: Check, that's indeed a lot easier. Added as `.cache`.

nutils/types.py
```diff
@@ -1484,6 +1484,55 @@ def del_lru(self):
       lru = self._sorted_keys.pop()
     del self[lru]
 
+def lru_cache(func=None, maxsize=None):
+  '''Buffer-aware LRU cache.
+
+  Return values are cached for immutable arguments are cached, otherwise
```
joostvanzwieten, line 1277: Typo: Return values -are cached- for immutable arguments are cached, ...

nutils/types.py
```diff
@@ -1484,6 +1484,55 @@ def del_lru(self):
...
+  caching is silently disabled. Any array arguments are equality-tested based
+  on the underlying buffer, with an callback placed on the memory release that
```
joostvanzwieten, line 1279: Typo: ..., with -an +a callback

nutils/types.py
```diff
@@ -1440,6 +1435,54 @@ def __init__(self, obj, flags=PyBUF_STRIDES):
...
+  def del_lru(self):
+    while self._sorted_keys:
+      lru = self._sorted_keys.pop()
+      if lru.count < self._last_sort: # not touched since last sort
+        break
+    else:
+      self._sorted_keys = sorted(self) # youngest first
+      self._last_sort = self.count
+      lru = self._sorted_keys.pop()
+    del self[lru]
```
joostvanzwieten, line 1272: Suggestion: implement [`clear`](https://docs.python.org/3.9/library/stdtypes.html#dict.clear): wipe `self._sorted_keys` and call super.

nutils/points.py
```diff
@@ -301,27 +300,27 @@ def masks(self):
     for pairs in self.duplicates:
       for i, j in pairs[1:]:
         masks[i][j] = False
-    return tuple(types.frozenarray(m, copy=False) for m in masks)
+    return types.tuple[numeric.freeze](masks)
```
joostvanzwieten, line 303: I'm not sure about using `types.tuple[f](s)` here either. It's hardly shorter than `tuple(map(f, s))`

nutils/topology.py
```diff
@@ -1627,7 +1626,7 @@ def connectivity(self):
           for (ichild, ichildedge), (jchild, jchildedge) in zip(self.basetopo.references[ielem].edgechildren[iedge], self.basetopo.references[jelem].edgechildren[jedge]):
             connectivity[offsets[ielem]+ichild][ichildedge] = offsets[jelem]+jchild
             connectivity[offsets[jelem]+jchild][jchildedge] = offsets[ielem]+ichild
-    return tuple(types.frozenarray(c, copy=False) for c in connectivity)
+    return types.tuple[numeric.freeze](connectivity)
```
joostvanzwieten, line 1629: See other comment about `types.tuple`.

nutils/types.py
gertjanvanzwieten: fixed

nutils/types.py
gertjanvanzwieten: fixed

nutils/types.py
gertjanvanzwieten: fixed

nutils/function.py
```diff
@@ -2793,8 +2793,8 @@ class PlainBasis(Basis):
   __slots__ = '_coeffs', '_dofs'
 
   def __init__(self, coefficients: Sequence[numpy.ndarray], dofs: Sequence[numpy.ndarray], ndofs: int, index: Array, coords: Array) -> None:
-    self._coeffs = tuple(map(types.frozenarray, coefficients))
-    self._dofs = tuple(map(types.frozenarray, dofs))
+    self._coeffs = tuple(types.frozenarray(c, dtype=float) for c in coefficients)
+    self._dofs = tuple(types.frozenarray(d, dtype=int) for d in dofs)
```
joostvanzwieten: I don't see why this is necessary. Especially since the last commit reverts the cast.

joostvanzwieten (changes requested):

joostvanzwieten:
I'll review the changes tomorrow.

nutils/elementseq.py
```diff
@@ -435,12 +435,13 @@ class _Take(References):
 
   __slots__ = 'parent', 'indices'
 
-  def __init__(self, parent: References, indices: numpy.ndarray) -> None:
-    _check_take(len(parent), indices)
-    assert len(indices) > 1, 'inefficient; this should have been `_Empty` or `_Uniform`'
+  @types.apply_annotations
+  def __init__(self, parent: types.strict[References], indices: types.arraydata) -> None:
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/pointsseq.py
```diff
@@ -450,12 +450,13 @@ class _Take(PointsSequence):
 
   __slots__ = 'parent', 'indices'
 
-  def __init__(self, parent, indices):
-    _check_take(len(parent), indices)
-    assert len(indices) > 1, 'inefficient; this should have been `_Empty` or `_Uniform`'
+  @types.apply_annotations
+  def __init__(self, parent, indices:types.arraydata):
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/function.py
```diff
@@ -683,8 +683,8 @@ def lower(self, *, coordinates: Tuple[evaluable.Array] = (), **kwargs: Any) -> e
 
 class _Elemwise(Array):
 
-  def __init__(self, data: Tuple[types.frozenarray, ...], index: Array, dtype: DType) -> None:
-    self._data = data
+  def __init__(self, data: Tuple, index: Array, dtype: DType) -> None:
+    self._data = tuple(map(types.arraydata, data))
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/function.py
```diff
@@ -2792,28 +2792,28 @@ class PlainBasis(Basis):
...
+    self._coeffs = tuple(types.arraydata(c) for c in coefficients)
+    self._dofs = tuple(map(types.arraydata, dofs))
     assert len(self._coeffs) == len(self._dofs)
     assert all(c.ndim == 1+coords.shape[0] for c in self._coeffs)
-    assert all(len(c) == len(d) for c, d in zip(self._coeffs, self._dofs))
+    assert all(c.shape[0] == d.shape[0] for c, d in zip(self._coeffs, self._dofs))
     super().__init__(ndofs, len(coefficients), index, coords)
 
-  def __getnewargs__(self) -> Tuple[Tuple[types.frozenarray], Tuple[types.frozenarray], int, Array, Array]:
+  def __getnewargs__(self) -> Tuple[Tuple, Tuple, int, Array, Array]:
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/function.py
```diff
@@ -2837,13 +2837,13 @@ class DiscontBasis(Basis):
...
-    self._coeffs = tuple(types.frozenarray(c, dtype=float) for c in coefficients)
+  def __init__(self, coefficients: Sequence, index: Array, coords: Array) -> None:
+    self._coeffs = tuple(types.arraydata(c) for c in coefficients)
     assert all(c.ndim == 1+coords.shape[0] for c in self._coeffs)
-    self._offsets = types.frozenarray(numpy.cumsum([0, *map(len, self._coeffs)]), copy=False)
+    self._offsets = numpy.cumsum([0] + [c.shape[0] for c in self._coeffs])
     super().__init__(self._offsets[-1], len(coefficients), index, coords)
 
-  def __getnewargs__(self) -> Tuple[Tuple[types.frozenarray], Array, Array]:
+  def __getnewargs__(self) -> Tuple[Tuple, Array, Array]:
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/types.py
```diff
@@ -1492,6 +1492,56 @@ def del_lru(self):
...
+        bases.append(obj)
+        key.append((Py_buffer(arg).buf, *arg.shape, *arg.strides, arg.dtype.str))
+      else:
+        key.append((type(arg), arg))
+    key = tuple(key)
+    try:
+      v = cache[key]
+    except KeyError:
+      v = cache[key] = func(*args)
+      assert _isimmutable(v)
```
soraros, line 1324: What about also add an argument `freeze=None` to `lru_cache` and have 
```python
freeze = lambda x: x if freeze is None else freeze
v = cached[key] = freeze(func(*args))
```
so that the functions in `numeric` can be left untouched. I think it's surprising that they return immutable results even if called with mutable arrays.
It would also be nice to expose an uncached version of `func` with `wrapper.nocache = func`.

nutils/types.py
```diff
@@ -1492,6 +1492,56 @@ def del_lru(self):
       lru = self._sorted_keys.pop()
     del self[lru]
 
+def lru_cache(func=None, maxsize=None):
```
soraros, line 1282: `frozenarray.lru` used to have `maxsize=128`, so caution needs be taken while moving from `frozenarray.lru`.
commit: https://github.com/evalf/nutils/pull/619/commits/19a51cbb506649aa91b64a3604edf27cce15c79f

nutils/types.py
```diff
@@ -1492,6 +1492,56 @@ def del_lru(self):
...
+  cache = {} if maxsize is None else lru_dict(maxsize)
+
+  @functools.wraps(func)
+  def wrapped(*args):
+    key = []
+    bases = []
+    for arg in args:
+      if isinstance(arg, numpy.ndarray):
+        obj = arg
+        while True:
```
soraros, line 1304: I guess we could also introduce a generator to traverse the bases in `numerics.py`
```python
def bases(obj, least=False):
  while isinstance(obj, numpy.ndarray):
    yield obj
    obj = obj.base
    if isinstance(obj, numpy.lib.stride_tricks.DummyArray):
      obj = obj.base
  if least:
    yield obj
```
Then simplify the code to
```python
obj = arg
for obj in bases(arg, least=True): # least = True so obj point to last nonnil after loop
  if isinstance(obj, arraydata): break
  if obj.flags.writeable:
    return func(*args)
```

Notice the iterator would also simplify `numeric.freeze`
```python
def freeze(arg):
  assert isinstance(arg, numpy.ndarray)
  for a in bases(arg):
    a.flags.writeable = False
  return arg
```

Appendix:
the base hierarchy in regular expression
```ocaml
let N = ndarray
    D = DummyArray
    W = arraydata
    N = None
in A(A|DA)*(W|N)
```

nutils/function.py
gertjanvanzwieten: This is necessary in relation to checking, rather than casting, the dtype in evaluable.Elemwise. The last commit no longer requires this as `dataclass` automatically converts data to the canonical type for its kind.

nutils/function.py
```diff
@@ -2943,14 +2943,14 @@ class StructuredBasis(Basis):
...
-    self._coeffs = tuple(tuple(types.frozenarray(a) for a in b) for b in coeffs)
-    self._start_dofs = tuple(map(types.frozenarray, start_dofs))
-    self._stop_dofs = tuple(map(types.frozenarray, stop_dofs))
-    self._dofs_shape = tuple(map(int, dofs_shape))
-    self._transforms_shape = tuple(map(int, transforms_shape))
+    self._coeffs = types.tuple[types.tuple[types.arraydata]](coeffs)
+    self._start_dofs = types.tuple[numeric.frozen](start_dofs)
+    self._stop_dofs = types.tuple[numeric.frozen](stop_dofs)
+    self._dofs_shape = types.tuple[int](dofs_shape)
+    self._transforms_shape = types.tuple[int](transforms_shape)
```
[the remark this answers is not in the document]
gertjanvanzwieten: fixed

nutils/points.py
gertjanvanzwieten: fixed

nutils/topology.py
gertjanvanzwieten: fixed

nutils/types.py
gertjanvanzwieten: We actually want to get rid of the polymorphic nature of lru-decorated functions, so in preparation to that it makes more sense to do the work on the function level where it's easier to follow. I added a line in the docstring indicating this future change.

nutils/types.py
gertjanvanzwieten: Even though cache entries are expected to self-destruct with the buffer, open ended caches are probably a bad idea. I changed the default back to 128.

nutils/types.py
gertjanvanzwieten: I had this decorator at one stage, but at present there are only two functions that traverse bases: `types._isimmutable`, which is recursive by nature, and `numeric.freeze` - so there is not that much gained by the additional abstraction. Moreover, since `numeric` imports `types`, the base iterator should be placed in the latter to be used by `_isimmutable`, which is an odd location. All in all I prefer to kick this particular can a little down the road.

gertjanvanzwieten: PS. I'm not sure that I understand the purpose of the regular expression.

soraros: I didn't notice `_isimutable` but here in `lru_cache` we also traverse the bases.
The regex is just for documentation purpose.

gertjanvanzwieten: Fair point. In the latest (and hopefully final) version I removed `numeric.freeze`, instead keeping for that purpose `types.frozenarray` which has the benefit of touching fewer code and being backwards compatible. Supporting `types.frozenarray`, `types._isimmutable` and `types.lru_cache` there is now `types._array_bases`.

joostvanzwieten (approved):

nutils/numeric.py
```diff
@@ -50,6 +50,16 @@ def overlapping(arr, axis=-1, n=2):
   overlapping.flags.writeable = False
   return overlapping
 
+def full(shape, fill_value, dtype):
+  'read-only equivalent to :func:`numpy.full`'
+
+  z = (0,)*len(shape)
+  f = numpy.ndarray(shape=shape, strides=z, dtype=dtype)
```
soraros, line 57: Could use `f = numpy.broadcast_to(fill_value, shape, dtype)` here.

gertjanvanzwieten: The point here is `strides`, which allows us to blow up a scalar to any shape without allocating more memory.

PR #633, opened 2021-06-23
add function.Custom (backport)

joostvanzwieten:
This is a backport of #628 to ease the transition from v6 to v7.

### DIFF
--- a/nutils/function.py
+++ b/nutils/function.py
@@ -251,6 +251,24 @@ class Points(Evaluable):
 
 POINTS = Points()
 
+class EvaluableConstant(Evaluable):
+  '''Evaluate to the given constant value.
+
+  Parameters
+  ----------
+  value
+      The return value of ``eval``.
+  '''
+
+  __slots__ = 'value'
+
+  def __init__(self, value):
+    self.value = value
+    super().__init__(())
+
+  def evalf(self):
+    return self.value
+
 class Tuple(Evaluable):
 
   __slots__ = 'items', 'indices'
--- a/nutils/function.py
+++ b/nutils/function.py
@@ -4075,6 +4075,81 @@ def insertaxis(arg, n, length):
   n = numeric.normdim(arg.ndim+1, n)
   return InsertAxis(arg, n, length)
 
+def broadcast_arrays(*arrays):
+  '''Broadcast the given arrays.
+
+  Parameters
+  ----------
+  *arrays : :class:`Array` or similar
+
+  Returns
+  -------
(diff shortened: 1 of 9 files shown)


nutils/function.py
```diff
@@ -3276,6 +3276,136 @@ def _diagonalize(self, axis, newaxis):
...
+  ...
+  >>> IntPower(asarray([2., 3.]), 2).eval()
+  array([[ 4.,  9.]])
+  >>> derivative(IntPower(a, 2), a).eval(a=numpy.array([2., 3.]))
+  array([[[ 4.,  0.],
+          [ 0.,  6.]]])
+  '''
+
+  def __init__(self, args, shape, dtype):
+    self._args = tuple(args)
```
gertjanvanzwieten, line 3366: `_args` overrides the attribute of the `Immutable` base class, which affects simplify amongst others. I suggest changing it to `__args`.

gertjanvanzwieten (changes requested):

nutils/function.py
```diff
@@ -3276,6 +3276,136 @@ def _diagonalize(self, axis, newaxis):
...
+    raise NotImplementedError('The partial derivative of {} to argument {} (counting from 0) is not defined.'.format(type(self).__name__, iarg)) # pragma: nocover
+
+  def _derivative(self, var, seen):
+    if self.dtype != float:
+      return super()._derivative(var, seen)
+    result = Zeros(self.shape + var.shape, dtype=self.dtype)
+    for iarg, arg in enumerate(self._args):
+      if not isinstance(arg, Array) or arg.dtype != float or var not in arg.dependencies and var != arg:
+        continue
+      d1 = asarray(self.partial_derivative(iarg, *self._args))
```
gertjanvanzwieten, line 3400: Let's add a shape check here to avoid obscure errors:
```
if d1.shape != self.shape + arg.shape:
  raise Exception('invalid partial derivative to argument #{}: expected shape {}, got {}'.format(iarg, self.shape + arg.shape, d1.shape))
```

gertjanvanzwieten: Tentatively added as e948305b.

nutils/function.py
gertjanvanzwieten: Tentatively added as 28a698f1.

joostvanzwieten: accepted

nutils/function.py
joostvanzwieten: Accepted with slight change in error message.

gertjanvanzwieten (approved):

PR #636, opened 2021-08-27
Simplifications

joostvanzwieten:


gertjanvanzwieten (approved):

nutils/evaluable.py
```diff
@@ -3236,6 +3236,40 @@ def _sum(self, axis):
...
+
+  def _rinflate(self, func, length, axis):
+    if equalindex(length, self._length):
+      return Ravel(Inflate(_inflate(func, self._ia, self._na, func.ndim - self.ndim), self._ib, self._nb))
+
+  def _intbounds_impl(self):
+    nbmin, nbmax = self._nb._intbounds
+    iamin, iamax = self._ia._intbounds
+    ibmin, ibmax = self._ib._intbounds
+    return iamin * nbmin + ibmin, iamax * nbmax + ibmax
```
joostvanzwieten, line 3336: If `iamax` is zero and `nbmax` is inf, then the upper bound is NaN. In this case the upper bound should be `ibmax`.

gertjanvanzwieten: fixed

nutils/evaluable.py
```diff
@@ -3236,6 +3236,41 @@ def _sum(self, axis):
...
+
+  def _rinflate(self, func, length, axis):
+    if equalindex(length, self._length):
+      return Ravel(Inflate(_inflate(func, self._ia, self._na, func.ndim - self.ndim), self._ib, self._nb))
+
+  def _intbounds_impl(self):
+    nbmin, nbmax = self._nb._intbounds
+    iamin, iamax = self._ia._intbounds
+    ibmin, ibmax = self._ib._intbounds
+    return (iamin and nbmin and iamin * nbmin) + ibmin, \
```
joostvanzwieten, line 3321: Since `nb` and `ia` are indices, their lower bounds are nonnegative and finite (or we'd run out of memory), hence `iamin * nbmin + ibmin` is always defined.

gertjanvanzwieten: fixed

joostvanzwieten: I think the stride is missing.

gertjanvanzwieten: indeed.. fixed

PR #637, opened 2021-08-30
Simplify Power(f,2) to Multiply(f,f) i/o Square(f)

gertjanvanzwieten:
This patch removes evaluable.Square in favour of Multiply, which has a more
powerful implementation for dealing with sparsity.


gertjanvanzwieten:
closed via #636

PR #653, opened 2021-10-21
derive function.Array from NDArrayOperatorsMixin

gertjanvanzwieten:


nutils/function.py
```diff
@@ -1109,6 +1025,15 @@ def _join_arguments(args_list: Iterable[Mapping[str, Argument]]) -> Dict[str, Ar
 
 # CONSTRUCTORS
 
+HANDLED_FUNCTIONS = {}
+
+def implements(np_function):
+  'Register an __array_function__ or __array_ufunc__ implementation for Array objects.'
```
joostvanzwieten, line 1033: Suggestion: use double bacticks for `__array_function__` and `__array_ufunc__`.

nutils/function.py
```diff
@@ -192,6 +192,16 @@ def lower(self, points_shape: _PointsShape, transform_chains: _TransformChainsMa
   def as_evaluable_array(self) -> evaluable.Array:
     return self.lower((), {}, {})
 
+  def __index__(self):
+    if self.arguments:
```
joostvanzwieten, line 196: I think this should be `if self.arguments or self.spaces:`.

nutils/sample.py
```diff
@@ -755,9 +753,8 @@ def eval_integrals_sparse(*integrals: evaluable.AsEvaluableArray, **arguments: M
   results : :class:`tuple` of arrays and/or :class:`nutils.matrix.Matrix` objects.
   '''
 
-  integrals = tuple(integral.as_evaluable_array.assparse for integral in integrals)
-  with evaluable.Tuple(tuple(integrals)).optimized_for_numpy.session(graphviz=graphviz) as eval:
-    return eval(**arguments)
+  warnings.deprecation('sample.eval_integrals_sparse is deprecated, use function.eval_sparse instead')
```
joostvanzwieten, line 756: `function.Array.eval` still uses `sample.eval_integrals`.

nutils/function.py
```diff
@@ -620,7 +620,8 @@ def lower(self, points_shape: _PointsShape, transform_chains: _TransformChainsMa
     coordinates = {space: evaluable.Transpose.to_end(evaluable.appendaxes(coords, add_points_shape), coords.ndim-1) for space, coords in coordinates.items()}
     return _CustomEvaluable(type(self).__name__, self.evalf, self.partial_derivative, args, self.shape[self._npointwise:], self.dtype, self.spaces, types.frozendict(self.arguments), points_shape, tuple(transform_chains.items()), tuple(coordinates.items()))
 
-  def evalf(self, *args: Any) -> numpy.ndarray:
+  @staticmethod
```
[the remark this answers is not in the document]
gertjanvanzwieten: done

nutils/function.py
gertjanvanzwieten: done

nutils/function.py
gertjanvanzwieten: fixed

nutils/sample.py
gertjanvanzwieten: ..which is not deprecated? Not sure that I see the issue. Anyway, I moved eval_sparse to evaluable now, and added function.eval in a separate commit.

nutils/function.py
```diff
@@ -494,9 +410,18 @@ class Custom(Array):
   pointwise axes and the shape of the result passed to :class:`Custom` should
   not include the pointwise axes.
 
+  For internal reasons, both ``evalf`` and ``partial_derivative`` must be
+  decorated as `classmethod` or `staticmethod`, meaning that they will not
```
joostvanzwieten, line 414: The single backticks should be double backticks.

nutils/sample.py
joostvanzwieten: Ah, I missed that `sample.eval_integrals` was not deprecated. Shouldn't we deprecated `sample.eval_integrals` as well?

nutils/function.py
gertjanvanzwieten: fixed

nutils/sample.py
gertjanvanzwieten: The thing is this (and Sample.integrate) is where _convert is called, so it's not the same as the new function.eval. It should go eventually but I don't think we can deprecate it yet.

joostvanzwieten: Ok.

joostvanzwieten (approved):

PR #675, opened 2022-04-19
Fix memory overflow on refinement

MartinEssink:
When the refinement level is high, the mask array rapidly increases in size. By only storing the indices higher refinement is possible.


gertjanvanzwieten:
I refined the concept a bit in branch refine-memory-fix of this repository, using union1d as a shorthand for concatenate and sort. If you can force push these commits to your branch we can merge the PR here, or I can open a new one if that's easier.

gertjanvanzwieten (approved):

PR #676, opened 2022-04-22
Minimum number of iterations in nonlinear solver

MartinEssink:
Added miniter to withsolve, to force a minimum number of iterations.


gertjanvanzwieten:
Thanks, this looks great. Could you just change the assert (miniter < maxiter) to raising a ValueError? We try to transition into using asserts only for internal consistency checks so that they can be safely disabled using the -O flag.

MartinEssink:
I have changed the assert to a ValueError.

gertjanvanzwieten (approved):

PR #682, opened 2022-05-18
Fields

gertjanvanzwieten:


nutils/expression_v2.py
```diff
@@ -709,6 +709,26 @@ def define_for(self, __name: str, *, gradient: Optional[str] = None, curl: Optio
                 raise ValueError('Cannot define the jacobian {!r}: dimension is negative.'.format(jacobian))
             setattr(self, jacobian, function.jacobian(geom, numpy.size(geom) - i))
 
+    def add_field(self, __names: tuple, __bases: tuple = (), shape: Tuple[int, ...] = (), dtype: function.DType = float):
```
joostvanzwieten, line 712: The annotation for `names` should be `Union[str, Sequence[str]]`. The annotation for `bases` is incorrect as well. Perhaps it's best to remove it.

examples/cylinderflow.py
```diff
@@ -58,32 +58,33 @@ def main(nelems: int, degree: int, reynolds: float, rotation: float, timestep: f
...
         domain.basis('spline', degree=(degree, degree-1), removedofs=((0,), None)),
         domain.basis('spline', degree=(degree-1, degree))]), J.T) / detJ
-    ns.pbasis = domain.basis('spline', degree=degree-1) / detJ
-    ns.u = function.dotarg('u', ns.ubasis)
-    ns.p = function.dotarg('p', ns.pbasis)
+    pbasis = domain.basis('spline', degree=degree-1) / detJ
+    ns.u = function.dotarg('u', ubasis)
+    ns.v = function.dotarg('v', ubasis)
+    ns.p = function.dotarg('p', pbasis)
+    ns.q = function.dotarg('q', pbasis)
```
joostvanzwieten, line 68: I think we can use `ns.add_field` here as well.

joostvanzwieten (approved):

joostvanzwieten:
Both driven cavity examples have a `function.dotarg` in `postprocess` that can be replaced with a `ns.add_field`.

nutils/expression_v2.py
gertjanvanzwieten: fixed (and turned bases into varargs)

examples/cylinderflow.py
gertjanvanzwieten: fixed

PR #701, opened 2022-08-23
reimplementation of command line interface

gertjanvanzwieten:


nutils/util.py
```diff
@@ -437,6 +438,44 @@ def binaryfile(path):
...
+    if not changed:
+        return f
+    sig = sig.replace(parameters=params)
+    def defaults_from_env(*args, **kwargs):
+        bound = sig.bind(*args, **kwargs)
+        bound.apply_defaults()
+        return f(*bound.args, **bound.kwargs)
+    defaults_from_env.__signature__ = sig
+    defaults_from_env.__wrapped__ = f
+    defaults_from_env.__doc__ = f.__doc__
```
joostvanzwieten, line 477: Why not `@functools.wraps(f)` decorate the wrapper? `@functools.wraps` copies more than `__wrapped__` and `__doc__`, for example: `__module__`, `__name__`, `__qualname__`.

nutils/util.py
```diff
@@ -437,6 +438,44 @@ def binaryfile(path):
     raise TypeError('binaryfile requires a path-like or file-like argument')
 
 
+def defaults_from_env(f):
+    '''Decorator for changing function defaults based on environment.
+
+    This decorator searches the environment for variables matching the pattern
+    ``NUTILS_MYPARAM``, where ``myparam`` is a parameter of the decorated
+    function. Only parameters with type annotation and a default value are
+    considered, and the string value is deserialized using stringly. In case
```
joostvanzwieten, line 449: Suggestion: Add a reference to the PyPI page of Stringly.

joostvanzwieten, line 449: Stringly capitalized?

nutils/util.py
```diff
@@ -512,6 +514,23 @@ def set_current(*args, **kwargs):
...
+    fmt = '%Y-%m-%d %H:%M:%S'
+    treelog.info(f'start {time.strftime(fmt,time.localtime(t0))}')
+    try:
+        yield
+    finally:
+        te = time.time()
+        m, s = divmod(int(te - t0), 60)
+        if m > 60:
+            m = '{}:{:02d}'.format(*divmod(m, 60))
+        treelog.info(f'finish {time.strftime(fmt,time.localtime(t0))}, elapsed {m}:{s:02d}')
```
joostvanzwieten, line 531: Alternatively you can use `datetime`. Advantage: you can write the time format of a `datetime` as format parameter:

```python
t0 = datetime.now()
treelog.info(f'start {t0:%Y-%m-%d %H:%M:%S}')
```

The difference between two `datetime` instances, `te - t0`, has a `total_seconds()` method.

nutils/cli.py
```diff
@@ -1,397 +1,361 @@
 """
-The cli (command line interface) module provides the `cli.run` function that
+he cli (command line interface) module provides the `cli.run` function that
```
joostvanzwieten, line 2: Typo.

nutils/cli.py
```diff
@@ -1,397 +1,361 @@
...
+        params.append(param)
+        cnames.append(param.name)
 
+    def in_context(*args, **kwargs):
+        with context(**{name: kwargs.pop(name) for name in cnames if name in kwargs}):
+            f(*args, **kwargs)
+
+    in_context.__signature__ = sig.replace(parameters=params)
+    in_context.__wrapped__ = f
+    in_context.__doc__ = f.__doc__
```
joostvanzwieten, line 48: Why not use `@functools.wraps(f)` here. See earlier comment.

nutils/util.py
gertjanvanzwieten: It does not introduce the new signature though. Better to combine wraps with assigning `__signature__` only?

joostvanzwieten: That's better IMO. If I'm not mistaken, Sphinx will ignore functions in a module if the `__module__` attribute doesn't match the global `__name__`.

gertjanvanzwieten: fixed

nutils/util.py
gertjanvanzwieten: fixed

nutils/cli.py
gertjanvanzwieten: fixed

nutils/cli.py
gertjanvanzwieten: fixed

joostvanzwieten (approved):

PR #708, opened 2022-09-01
add Pointwise.return_type

gertjanvanzwieten:
This patch adds the return_type class method to evaluable.Pointwise which
defines the output dtype for given input dtypes. This replaces the constructor
logic that evaluated evalf in numpy.ones, which is problematic as one may not
be in the evaluable range (e.g. arctanh). Moreover, the new method allows
evaluable to deviate from Numpy's logic. Finally, the new function replaces all
constructor definitions that served only to check input types (e.g. Greater,
Less).


nutils/evaluable.py
```diff
@@ -2534,11 +2543,7 @@ def _simplified(self):
 
 class Real(Pointwise):
     __slots__ = ()
-
-    @types.apply_annotations
-    def __init__(self, arg: asarray):
-        assert arg.dtype == complex
-        super().__init__(arg)
+    return_type = lambda T: float if T in (float, complex) else int
```
joostvanzwieten, line 2546: `numpy.real(numpy.array([False])).dtype` gives `dtype('bool')`

nutils/evaluable.py
```diff
@@ -2552,11 +2557,7 @@ def _simplified(self):
 
 class Imag(Pointwise):
     __slots__ = ()
-
-    @types.apply_annotations
-    def __init__(self, arg: asarray):
-        assert arg.dtype == complex
-        super().__init__(arg)
+    return_type = lambda T: float if T in (float, complex) else int
```
joostvanzwieten, line 2560: `numpy.imag(numpy.array([False])).dtype` gives `dtype('bool')`

nutils/evaluable.py
gertjanvanzwieten: changed to `return_type = lambda T: float if T == complex else T`

nutils/evaluable.py
gertjanvanzwieten: idem

joostvanzwieten (approved):

PR #714, opened 2022-09-07
Cli

gertjanvanzwieten:


CHANGELOG
```diff
@@ -5,6 +5,28 @@ The following overview lists user facing changes as well as newly added
 features in inverse chronological order.
 
 
+REMOVED: nutils configuration file
+
+Suport for the nutils configuration file (which used to be located in either
```
joostvanzwieten, line 10: Typo: `Suport` -> `Support`

CHANGELOG
```diff
@@ -5,6 +5,28 @@ The following overview lists user facing changes as well as newly added
 features in inverse chronological order.
 
 
+REMOVED: nutils configuration file
```
joostvanzwieten, line 8: Capitalize Nutils? (There are three occurrences in the CHANGLOG patch.)

nutils/util.py
```diff
@@ -777,19 +777,20 @@ def name_of_main():
...
 @contextlib.contextmanager
 @defaults_from_env
-def add_htmllog(outrootdir: str = '~/public_html', outrooturi: str = '', scriptname: str = os.path.basename(sys.argv[0]), outdir: str = '', outuri: str = ''):
+def add_htmllog(outrootdir: str = '~/public_html', outrooturi: str = '', scriptname: str = '', outdir: str = '', outuri: str = ''):
     '''Context to add a HtmlLog to the active logger.'''
 
     import html, base64, bottombar
 
+    if not scriptname and (not outdir or outrooturi and not outuri):
+        scriptname = name_of_main()
```
joostvanzwieten, line 786: What's the reason for replacing `sys.argv[0]` with `__main__.__name__` (via `util.name_of_main`)?

gertjanvanzwieten: You main as a default value? The relevant attributes of __main__ are not initialized until all the other modules are loaded, unfortunately.

CHANGELOG
gertjanvanzwieten: fixed

CHANGELOG
gertjanvanzwieten: fixed

nutils/util.py
joostvanzwieten: If I understand the patch correctly, there are two changes:

1. Set the default value of parameter `scriptname` to the empty string and conditionally replace an empty `scriptname` with `os.path.basename(sys.argv[0])` in the function body.
2. Replace `os.path.basename(sys.argv[0])` with `util.name_of_main()`.

I understand that change 2 requires change 1. But why change 2?

gertjanvanzwieten: Ah, that's for scenarios such as python -m, or python -m pdb -m. I could not think of a robust way of distilling the application name from argv covering all potential situations. In any case argv[0] is too simplistic.

joostvanzwieten: Ok. Can you add this to the commit message?

gertjanvanzwieten: done

PR #734, opened 2022-12-01
serialized evalf

joostvanzwieten:
`evaluable.Loop*.evalf` repeatedly loops over all ops and calls method `evalf()` for each op. This PR speeds-up this process by making as many `evalf` methods static, e.g. by assigning a Numpy function to `evalf`, and moving the lookup of the `evalf` attribute outside the loop.

### DIFF
--- a/nutils/evaluable.py
+++ b/nutils/evaluable.py
@@ -3918,7 +3918,7 @@ class _LoopIndex(Argument):
 
 class LoopSum(Array):
 
-    __cache__ = '_serialized'
+    __cache__ = '_serialized_loop'
 
     def prepare_funcdata(arg):
         # separate shape from array to make it simplifiable (annotations are
@@ -3941,12 +3941,12 @@ class LoopSum(Array):
         super().__init__(args=(shape, length, *self._invariants), shape=self.func.shape, dtype=self.func.dtype)
 
     @property
-    def _serialized(self):
+    def _serialized_loop(self):
         indices = {d: i for i, d in enumerate(itertools.chain([self.index], self._invariants, self._dependencies))}
         return tuple((dep, tuple(map(indices.__getitem__, dep._Evaluable__args))) for dep in self._dependencies)
 
     def evalf(self, shape, length, *args):
-        serialized = self._serialized
+        serialized = self._serialized_loop
         result = numpy.zeros(shape, self.dtype)
         for index in range(length):
             values = [numpy.array(index)]
@@ -3956,7 +3956,7 @@ class LoopSum(Array):
         return result
 
     def evalf_withtimes(self, times, shape, length, *args):
-        serialized = self._serialized
+        serialized = self._serialized_loop
         subtimes = times.setdefault(self, collections.defaultdict(_Stats))
         result = numpy.zeros(shape, self.dtype)
         for index in range(length):
@@ -4142,7 +4142,7 @@ class LoopConcatenate(Array):
 
 class LoopConcatenateCombined(Evaluable):
 
-    __cache__ = '_serialized'
+    __cache__ = '_serialized_loop'
 
     @types.apply_annotations
     def __init__(self, funcdatas: types.tuple[asarrays], index_name: types.strictstr, length: asindex):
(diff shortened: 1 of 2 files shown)


nutils/evaluable.py
```diff
@@ -1567,8 +1574,7 @@ def _simplified(self):
         if equalindex(self.func.shape[-1], 1):
             return reciprocal(self.func)
 
-    def evalf(self, arr):
-        return numeric.inv(arr)
+    evalf = staticmethod(numeric.inv)
```
gertjanvanzwieten, line 1577: I haven't checked, but are you sure staticmethod is necessary here? I thought this is implied if the object is not of type function. (same in other similar places)

nutils/evaluable.py
```diff
@@ -2285,13 +2287,13 @@ def _unravel(self, axis, shape):
 
 class Reciprocal(Pointwise):
     __slots__ = ()
-    evalf = numpy.reciprocal
+    evalf = staticmethod(numpy.reciprocal)
```
gertjanvanzwieten, line 2290: see previous comment; why are we changing this?

nutils/types.py
```diff
@@ -349,7 +349,6 @@ def __get__(self, instance, owner):
             cached_value = getattr(instance, self.cache_attr)
         except AttributeError:
             value = self.fget(instance)
-            assert _isimmutable(value)
```
gertjanvanzwieten, line 352: can you explain in the commit message why this condition can be relaxed? I don't immediately see it

nutils/evaluable.py
```diff
@@ -4164,14 +4168,18 @@ def _serialized(self):
         indices = {d: i for i, d in enumerate(itertools.chain([self._index], self._invariants, self._dependencies))}
         return tuple((dep, tuple(map(indices.__getitem__, dep._Evaluable__args))) for dep in self._dependencies)
 
+    @property
+    def _serialized_evalf(self):
+        return tuple((dep.evalf, indices) for dep, indices in self._serialized)
```
gertjanvanzwieten, line 4173: We are propagating the situation here of two identical `_serialized_evalf` implementations mirrorring near-identical implementations of `_serialized`. Could we consider an index-holding base class here for LoopSum and LoopConcatenateCombined?

gertjanvanzwieten (approved):

nutils/evaluable.py
joostvanzwieten: AFAIK a function `func` defined with `def` implements the descriptor protocol such that when calling `func.__get__` with a not-None instance the result is a method call bound to that instance (comparable to `functools.partial(func, instance)`). It doesn't matter if `func` is defined as part of a class or not.

Numpy ufuncs don't implement the descriptor protocol, so wrapping those in a `staticmethod` is unnecessary. I added `staticmethod` for symmetry, but if you don't like it, I'll strip all `staticmethod` from Numpy ufuncs.

gertjanvanzwieten: Right, no I guess there are merits to being explicit, I just wasn't sure if there were reasons beyond that somehow relating to the subsequent commits, especially when I noticed you revisiting the existing Pointwise objects. Let's keep the changes you made.

nutils/types.py
gertjanvanzwieten: Clear. It seems to me that option 3 is doable if we recognize bound methods (using inspect.isinstance) and testing immutability of `__self__`, functions (for staticmethods) and numpy ufuncs as a third special category. What do you think?

joostvanzwieten: This seems to work: `isinstance(value, numpy.ufunc)`. I'll try that.

joostvanzwieten: We have decided to go for option 2 after all, as it appears to be difficult to cover all possible values of `evalf`, in particular instances of external classes that define a `__call__` method (e.g. `nutils_poly.MulPlan` which will be introduced in #700).

nutils/evaluable.py
joostvanzwieten: I'd like the idea of merging common parts of `LoopSum` and `LoopConcatenateCombined`, but propose to defer this to a separate PR. I'd like to finish #700 first.

nutils/types.py
joostvanzwieten: A side effect of using `cached_property` is that we need a `__dict__`. Not a big problem IMO. We could even consider dropping `__slots__`.

gertjanvanzwieten: Agreed. I suspect there was never a strong reason to use those, and probably less so today with developments such as https://github.com/python/cpython/pull/28802

gertjanvanzwieten: I just noticed that currently you keep `__slots__` and add `__dict__`. Better drop `__slots__` I would say, even if we decide to keep the remaining ones for now.

gertjanvanzwieten: PS when we do ultimately drop all slots I think we can consider replacing all of `__cache__` by `cached_property` decorators.

gertjanvanzwieten (changes requested):
Other than the slots=dict construction it looks good to me.

gertjanvanzwieten (approved):

Issue #738, opened 2022-12-02
make `Evaluable` array-only


joostvanzwieten:
To simplify the transition of the evaluable module to Rust we should let `Evaluable` be array-only.

To be removed or rewritten as an `Array`:

- [x] `EVALARGS` (#859)
- [x] `EvaluableConstant` (#847)
- [ ] `Tuple`
- [x] `SparseArray` (#787)
- [ ] `Eig`
- [x] `EvaluableTransformChain` (#739)
- [x] `Points` and `NPoints` (#846)

Open question:

- [ ] How to handle functions that have multiple outputs like `eig`?

joostvanzwieten:
`evaluable.Evaluable` subclasses other than `evaluable.Array` are used only within the `evaluable` module, so I consider this issue completed.

PR #739, opened 2022-12-02
array-valued evaluable coordinate transformations

joostvanzwieten:
In an effort to make the `evaluable` module array-only (#738), this PR replaces the `EvaluableTransform*` family of classes with array-valued evaluable coordinate transformations.


nutils/evaluable.py
```diff
@@ -3900,6 +3900,175 @@ def _intbounds_impl(self):
...
+
+class TransformBasis(Array):
+    '''Extended basis of a source coordinate system in reference coordinates
+
+    The basis (the columns of this matrix) spans the space of reference
+    coordinates (spatial part). If the source coordinate system has a lower
+    dimension, say `n`, than the reference coordinate system, the first `n`
+    basis vectors span the space of source coordinates mapped to the reference
+    for the given source index. The remainder is *not* a span of the complement
+    space.
```
gertjanvanzwieten, line 4031: I think I would understand this better if it was introduced along the lines of "given a transformation chain into n-dimensional space, the TransformBasis represents a vector basis of dimension n with the special property that columns 1 .. m < n form a basis for any m-dimensional subspace traversed by the transformation chain". I'm also not sure about calling it an "Extended basis" which is not very informative. Perhaps "cascaded vector basis for a transformation chain" connects better with the suggested paragraph.

gertjanvanzwieten (approved):

nutils/evaluable.py
joostvanzwieten: Let's drop extended indeed. I'm not sure if we should mention transformation chains. That's an implementation detail. I'd prefer to talk about coordinate systems and mappings from one to another. Let's discuss this next week.

gertjanvanzwieten (approved):

PR #752, opened 2022-12-10
remove EvaluationError

gertjanvanzwieten:
Remove the evaluable.EvaluationError exception type, which presented debugging difficulties by shielding the actual point of failure from PDB. In the new situation, the evaluation stack is logged at error level prior to reraising the original exception.

### DIFF
--- a/nutils/evaluable.py
+++ b/nutils/evaluable.py
@@ -2733,7 +2733,8 @@ class Sampled(Array):
 
     @staticmethod
     def evalf(points, expect):
-        assert numpy.equal(points, expect).all(), 'illegal point set'
+        if points.shape != expect.shape or not numpy.equal(points, expect).all():
+            raise ValueError('points do not correspond to original sample')
         return numpy.eye(len(points))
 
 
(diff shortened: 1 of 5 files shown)


nutils/evaluable.py
```diff
@@ -422,12 +422,10 @@ def eval(**args):
             log.info('total time: {:.0f}ms\n'.format(tottime/1e6) + '\n'.join('{:4.0f} {} ({} calls, avg {:.3f} per call)'.format(t / 1e6, k, n, t / (1e6*n))
                                                                               for k, t, n in sorted(aggstats, reverse=True, key=lambda item: item[1]) if n))
 
-    def _stack(self, values):
-        lines = ['  %0 = EVALARGS']
-        for (op, indices), v in zip(self.serialized, values):
-            lines[-1] += ' --> ' + type(v).__name__
-            if numeric.isarray(v):
-                lines[-1] += '({})'.format(','.join(map(str, v.shape)))
+    @property
```
joostvanzwieten, line 427: I'm not sure if a property makes sense for iterators. It's not immediately clear why you would get a new iterator every time you access the `_iter_stack` attribute for some instance of `Evaluable`.

joostvanzwieten (approved):

nutils/evaluable.py
gertjanvanzwieten: true, fixed

joostvanzwieten (approved):

PR #758, opened 2022-12-16
Remove apply_annotation from transform module

gertjanvanzwieten:


nutils/transform.py
```diff
@@ -85,8 +86,9 @@ class TransformItem(types.Singleton):
 
     __slots__ = 'todims', 'fromdims'
 
-    @types.apply_annotations
-    def __init__(self, todims, fromdims: int):
+    def __init__(self, todims: Integral, fromdims: Integral):
```
joostvanzwieten, line 89: Why `Integral` and not `int`?

nutils/transformseq.py
```diff
@@ -606,7 +623,11 @@ def index_with_tail(self, trans):
         return flatindex, tail
 
 
-class MaskedTransforms(Transforms):
+def MaskedTransforms(parent: Transforms, indices):
```
joostvanzwieten, line 626: We're not calling `MaskedTransforms` outside this module, so I'm fine with omitting this public function altogether.

nutils/transformseq.py
```diff
@@ -647,7 +668,11 @@ def index_with_tail(self, trans):
             return int(index), tail
 
 
-class ReorderedTransforms(Transforms):
+def ReorderedTransforms(parent: Transforms, indices):
```
joostvanzwieten, line 671: Same as `MaskedTransforms`.

nutils/transform.py
gertjanvanzwieten: By using `Integral` we allow `numpy.int_` in addition to `int`, which would otherwise require hunting down and preprocessing all arguments that originate from a numpy operation. The hash of both types corresponds so for `Singleton`'s purpose it is fine to take values as they come.

nutils/transformseq.py
gertjanvanzwieten: Fixed. For uniformity I now also reverted `PlainTransforms` and added `transform.iscanonical` to test if the argument is properly preprocessed. This is perhaps a bit controversial - the unit tests are passing with calling `transform.canonical` in most places but I'm not sure if that will be correct in all scenarios. This in turn made me wonder if we actually use the `Singleton` nature for the sequences? If not we can simply process arguments in the constructor.

nutils/transformseq.py
gertjanvanzwieten: Fixed.

nutils/transform.py
joostvanzwieten: Ok.

tests/test_transformseq.py
```diff
@@ -400,47 +400,48 @@ def test_invalid_dimensions(self):
...
     def test_PlainTransforms_invalid_fromdims(self):
         with self.assertRaisesRegex(ValueError, 'expected transforms with fromdims=2, but got .*'):
-            nutils.transformseq.PlainTransforms([(x2, e00), (x2, e11)], 2, 2)
+            nutils.transformseq.PlainTransforms(((x2, e00), (x2, e11)), 2, 2)
 
     def test_PlainTransforms_multiple_fromdims(self):
         with self.assertRaisesRegex(ValueError, 'expected transforms with fromdims=2, but got .*'):
-            nutils.transformseq.PlainTransforms([(x2, c00, e00), (x2, c01, i20, i20)], 2, 2)
+            nutils.transformseq.PlainTransforms(tuple(transform.canonical((x2, *tail)) for tail in ((c00, e00), (c01, i20, i20))), 2, 2)
+            nutils.transformseq.PlainTransforms(tuple(transform.canonical((x2, *tail)) for tail in ((c00, e00), (c01, i20, i20))), 2, 2)
```
joostvanzwieten, line 412: I don't think we every reach the second line.

joostvanzwieten (changes requested):

nutils/transformseq.py
joostvanzwieten: I'm not sure if we need `Singleton` either, but changing `Singleton` into `Immutable` warrants a separate commit and is a bit outside the purpose of this PR.

tests/test_transformseq.py
gertjanvanzwieten: Reverted all changes to transformsseq in light of the plan to make sequences non-singleton.

nutils/transformseq.py
gertjanvanzwieten: We'll open another PR for this.

joostvanzwieten (approved):

PR #759, opened 2022-12-18
Fix unit tests failing with Numpy 1.24.0

gertjanvanzwieten:
This PR aims to fix the unit tests that broke with the release of Numpy version 1.24.0 earlier today (though no mention of it on the website as yet). The failing workflow of the master branch is https://github.com/evalf/nutils/actions/runs/3714405319.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -96,7 +96,7 @@ jobs:
           python -um pip install "$_wheel[import_gmsh]"
       - name: Install Scipy
         if: ${{ matrix.matrix-backend == 'scipy' }}
-        run: python -um pip install --upgrade --upgrade-strategy eager scipy
+        run: python -um pip install --upgrade scipy
       - name: Configure MKL
         if: ${{ matrix.matrix-backend == 'mkl' }}
         run: |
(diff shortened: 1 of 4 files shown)


joostvanzwieten (approved):

gertjanvanzwieten:
Lowering precision by one decimal did not cut it for `test_inv`; I lowered it by two now but that is becoming hard to justify. Meanwhile I have not been able to reproduce any of this locally so I have no idea what is going on. Even the fix for replacing `numpy.bool` (the only problem I managed to reproduce and fix locally) by `numpy.bool_` gives a new error here. In light of all this weirdness, I wonder if we should update dependencies to avoid Numpy 24 for now and reassess after the first point release or so.

gertjanvanzwieten:
So.. even two orders is not enough. Aside from the fact that it is probably better to specify accuracies per test, rather than one to fit all as we do now, dropping three orders for the inverse of a square is definitely not something to ignore. The question is what we do in the meantime pending investigations (which are complex, given that this is not locally reproduceable).

joostvanzwieten (approved):

PR #761, opened 2023-01-10
Remove apply_annotation from topology module

gertjanvanzwieten:


nutils/mesh.py
```diff
@@ -53,7 +53,7 @@ def rectilinear(richshape: Sequence[Union[int, Sequence[float]]], periodic: Sequ
 _oldrectilinear = rectilinear  # reference for internal unittests
 
 
-def line(nodes: Union[int, Sequence[float]], periodic: bool = False, bnames: Optional[Sequence[Tuple[str, str]]] = None, *, name: Optional[str] = None, space: str = 'X', root: Optional[TransformItem] = None) -> Tuple[Topology, function.Array]:
```
joostvanzwieten, line 56: For a line `bnames` is not a sequence of tuples.

joostvanzwieten: This patch actually fixes this. Nevermind!

PR #765, opened 2023-01-19
fix basis multipatch topology

joostvanzwieten:
This PR fixes creating a basis on a `MultipatchTopology` with C^0 continuity at the patch interfaces. In addition, this PR deduplicates code for merging dofs.

### DIFF
--- a/nutils/_util.py
+++ b/nutils/_util.py
@@ -9,6 +9,7 @@ import os
 import numpy
 import collections.abc
 import inspect
+import itertools
 import functools
 import operator
 import numbers
@@ -18,6 +19,7 @@ import io
 import contextlib
 import treelog
 import datetime
+from typing import Iterable, Sequence, Tuple
 
 supports_outdirfd = os.open in os.supports_dir_fd and os.listdir in os.supports_fd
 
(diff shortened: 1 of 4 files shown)


nutils/_util.py
```diff
@@ -766,4 +768,52 @@ def cli(f, *, argv=None):
...
+            index = parent
+            parent = index_map[index]
+        return index
+    for merge_set in merge_sets:
+        resolved = list(map(resolve, merge_set))
+        index_map[resolved] = min(resolved)
+    new_indices = itertools.count()
+    for iin, ptr in enumerate(index_map):
+        index_map[iin] = next(new_indices) if iin == ptr else index_map[ptr]
+    return index_map, next(new_indices)
```
gertjanvanzwieten, line 816: Less elegant, but could `numpy.unique` be actually faster? together with the previous comment that would amount to:

```python
index_map = numpy.arange(nin)
for merge_set in merge_sets:
    resolved = list(merge_set)
    resolved.extend(index_map[resolved])
    index_map[resolved] = min(resolved)
unique, index = numpy.unique(index_map, return_inverse=True)
return index, len(unique)
```

nutils/topology.py
```diff
@@ -1647,29 +1647,16 @@ def _basis_c0_structured(self, name, degree):
...
+        def pairs():
+            for ielem, ioppelems in enumerate(self.connectivity):
+                for iedge, jelem in enumerate(ioppelems):  # loop over element neighbors and merge dofs
+                    if jelem < ielem:
+                        continue  # either there is no neighbor along iedge or situation will be inspected from the other side
+                    jedge = util.index(self.connectivity[jelem], ielem)
+                    idofs = offsets[ielem] + self.references[ielem].get_edge_dofs(degree, iedge)
+                    jdofs = offsets[jelem] + self.references[jelem].get_edge_dofs(degree, jedge)
+                    yield from zip(idofs, jdofs)
+        dofmap, ndofs = util.merge_index_map(offsets[-1], pairs())
```
gertjanvanzwieten, line 1659: I'd be in favour of replacing the `pairs` method by a generator comprehension:

```python
# To merge matching dofs we loop over the connectivity table to find
# neighbouring elements (limited to jelem > ielem to consider every
# neighbour pair exactly once as well as ignore exterior boundaries)
# and mark the degrees of freedom on both sides to be equal.
dofmap, ndofs = util.merge_index_map(offsets[-1], (pair
    for ielem, ioppelems in enumerate(self.connectivity)
        for iedge, jelem in enumerate(ioppelems) if jelem > ielem
            for pair in zip(
                offsets[ielem] + self.references[ielem].get_edge_dofs(degree, iedge),
                offsets[jelem] + self.references[jelem].get_edge_dofs(degree, util.index(self.connectivity[jelem], ielem)))))
```

nutils/_util.py
```diff
@@ -766,4 +768,52 @@ def cli(f, *, argv=None):
...
+
+    index_map = numpy.arange(nin)
+    def resolve(index):
+        parent = index_map[index]
+        while index != parent:
+            index = parent
+            parent = index_map[index]
+        return index
+    for merge_set in merge_sets:
+        resolved = list(map(resolve, merge_set))
```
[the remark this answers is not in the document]
joostvanzwieten: You need to resolve the pointer until the end. Try this one: `util.merge_index_map(4, [[2, 3], [1, 2], [0, 3]])` (unittest updated).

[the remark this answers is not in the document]
gertjanvanzwieten: Clear, ok so we leave things as they are.

nutils/_util.py
gertjanvanzwieten: This also doesn't apply anymore then.

nutils/topology.py
gertjanvanzwieten: Actually, the original said to continue if `jelem < ielem`, so the inverse of that would be `jelem >= ielem` - I just thought that `jelem == ielem` was irrelevant. Would it be as simple as that? Also, I'm not sure if you intentionally didn't copy the comment block, but I thought that would be useful to describe what happens in the triply nested loop now that we cannot place comments inline anymore.

joostvanzwieten: Strange, I really thought I had copied the comment as well. I've pushed a version with comment and `jelem >= ielem`.

The condition `jelem >= ielem` is not required. It only speeds things up.

gertjanvanzwieten (approved):

PR #768, opened 2023-01-20
Normal

gertjanvanzwieten:
Generalize function.normal so that it can be used to evaluate the boundary normal of manifolds in higher dimensional space.


nutils/evaluable.py
```diff
@@ -1094,28 +1094,29 @@ def evalf(evalargs):
         return weights
 
 
-class Normal(Array):
+class Orthonormal(Array):
     'normal'
```
joostvanzwieten, line 1099: I think we can improve a bit on the documentation. Something like: Orthonormalize the vector to/w.r.t the basis.

nutils/evaluable.py
```diff
@@ -1094,28 +1094,29 @@ def evalf(evalargs):
...
+        self._vector = vector
+        super().__init__(args=(basis, vector), shape=vector.shape, dtype=float)
 
     def _simplified(self):
-        if _equals_scalar_constant(self.shape[-1], 1):
-            return Sign(Take(self.lgrad, constant(0)))
-        unaligned, where = unalign(self.lgrad, naxes=self.ndim - 1)
+        if _equals_scalar_constant(self._basis.shape[-1], 0):
+            # basis is empty so we can simply normalize vector
+            return einsum('Ij,I->Ij', self._vector, power(einsum('Ij,Ij->I', self._vector, self._vector), -.5))
```
joostvanzwieten, line 1114: What do you think of separating `Normalize` from `Orthonomal` (`Orthogonal(ize)`)? We now have two different paths for normalization: this one and `numeric.normalize` in `evalf`. The two options are a dedicated class `Normalize` with `Normalize.evalf = numeric.normalize`, or a function `normalize` that returns the `einsum(..., vector, power(...))`.

nutils/evaluable.py
```diff
@@ -1115,13 +1116,26 @@ def _simplified(self):
...
+            GG = numpy.einsum('...ki,...kj->...ij', G, G)
+            v1 = numpy.einsum('...ij,...i->...j', G, n)
+            v2 = numpy.linalg.solve(GG, v1)
+            v3 = numpy.einsum('...ij,...j->...i', G, v2)
+            return numeric.normalize(n - v3)
+
+    else:
+
+        # Faster implementation for Numpy versions that support QR
+        # factorization of arrays with arbitrarily many dimensions.
```
joostvanzwieten, line 1132: With QR we don't get an exception when `G` is linearly dependent. I'm not sure if we want this.

nutils/evaluable.py
gertjanvanzwieten: Ok.. I was going to argue that combining the two has the large benefit of having a very simple derivative (which it does) but in the process of demonstrating this point I came to realize that this derivative **requires** a tangent space of dimension nd-1, and so is incorrect for the generalized case introduced in this PR. Here are the relevant derivations (which I'm going to leave in comments this time):

```
P := I - G (G^T G)^-1 G^T (orthogonal projector)
P' = Q P + P Q^T (derivative) where Q := -G (G^T G)^-1 G'^T

  * P^T = P (symmetric)
  * P P = P (projection)
  * (I - P) P' = Q P (useful identity)

n := P v (orthogonal projection of v)
n' = P' v + P v' (derivative)

N := n / sqrt(n^T n) (orthonormal projection of v)
N' = (I - N N^T) n' / sqrt(n^T n) (derivative)
```

Now, iff the kernel of `G` has dimension 1, then it has a basis in `N`, hence `P = N N^T`, and the derivative of `N` simplifies to `N' = Q N`. Note that this doesn't even contain the derivative of `v` anymore, which makes sense when you realize that `v` only contributes an orientation, which is discrete (just stating it here because it took me a while again to have the picture complete in my head).

Back to your suggestion. If we keep the current `Orthonormal`, then we can simply return the long form of `N'` if the kernel dimenson is >= 2. If we separate the two operations, then we will need to recognize during derivative taking that the argument of `Normalize` is `Orthogonalize` and that the kernel is of dim 1, which is a bit messy but not hard. I'm not sure I see the purpose though: for me the main reason for `Orthonormal` to exist as a dedicated operation is the derivative, which could never result from a generic implementation. This does not apply to either `Orthogonalize` or `Normalize`, which will do essentially the same as their generic counterparts, so I think we're better off implementing them as such.

gertjanvanzwieten: Having thought about it a bit more, I propose we revert `Orthonormal` to being specific for one-dimensional kernels, and lower other situations directly to the generic expression so we don't have to write out the lengthy derivative for these cases.

nutils/evaluable.py
gertjanvanzwieten: changed it to "make a vector orthonormal to a subspace"

nutils/evaluable.py
gertjanvanzwieten: I reworked the general derivative to `N' = Q N + (P - N N^T) (Q^T v + v') / |n|`, in which the first term represents the special case for 1D kernels, and the second term contains two self references, so in the end I think there is merit to making `Orthonormal` generic and implementing the derivative manually. Derivations are in the comments.

nutils/evaluable.py
gertjanvanzwieten: Removed it.

joostvanzwieten (approved):

PR #774, opened 2023-02-04
Linearize

gertjanvanzwieten:


tests/test_function.py
```diff
@@ -1403,3 +1403,15 @@ def test_divide(self):
         f = function.Argument('test', shape=(2, 3), dtype=int)
         self.assertIsNot(f / 1, f)
         self.assertIsNot(f / 1., f)
+
+
+class linearize(TestCase):
+
+    def test(self):
+        f = function.linearize(function.Argument('u', shape=(3, 4), dtype=float)**3
```
joostvanzwieten, line 1411: Can you maybe add a scalar argument to the test? We are relying on `numpy.sum(x, ())` returning `x` unmodified, but the documentation is not super clear on this (an empty tuple is not necessarily a tuple of ints).

joostvanzwieten (approved):

tests/test_function.py
gertjanvanzwieten: I changed p into a scalar - anyhow cleaner that way

joostvanzwieten (approved):

PR #784, opened 2023-04-06
remove CacheMeta-based caching

gertjanvanzwieten:
This PR removes all uses of the `__cache__` mechanism, facilitated by the `CacheMeta` meta class, and replaces it by the cached_property and `lru_cache` decorators which are considered to give better visibility to caching structures. The `CacheMeta` class is removed as no longer relevant.


gertjanvanzwieten:
Oh there is actually an unresolved problem: I had to remove the base properties `Reference.vertices` and `Reference.edge_vertices` because they cannot otherwise be overridden in `MosaicReference`. @joostvanzwieten Would it be acceptable to use an annotation like `vertices: numpy.ndarray` instead?

nutils/element.py
```diff
@@ -24,10 +25,19 @@
...
 
 class Reference(types.Singleton):
-    'reference element'
+    '''Reference element.
 
-    __slots__ = 'ndims',
-    __cache__ = 'connectivity', 'edgechildren', 'volume', 'centroid', '_linear_bernstein', 'getpoints'
+    .. py:attribute:: vertices
+
+       Array of shape `nverts x ndims` that defines the coordinates of the
```
joostvanzwieten, line 32: Code should be enclosed in double backticks.

nutils/element.py
```diff
@@ -24,10 +25,19 @@
...
+       element's vertices.
+
+    .. py:attribute:: edge_vertices
+
+       Sequence of integer arrays that specifies per edge (outer sequence,
+       corresponding to `edges`) for each vertex (inner sequence, corresponding
```
joostvanzwieten, line 38: double backticks

nutils/element.py
```diff
@@ -24,10 +25,19 @@
...
+       to `edges[iedge].vertices`) its index in `vertices`.
```
joostvanzwieten, line 39: double backticks

joostvanzwieten (approved):

joostvanzwieten (approved):

PR #788, opened 2023-05-09
Variable arguments for Add and Multiply

gertjanvanzwieten:
This PR changes the Multiply and Add operations to support any number of arguments for faster evaluation and more efficient simplifications.

For operations organized in a tree such as `(A + B) + (C + D)` it is difficult to discover relevant structures, for instance `A` and `D` having matching sparsity. In the current form, the only way to properly discover this is if `(A + B)._add(C + D)` dispatches to `A._add(C + D)` as well as the reverse `(C + D)._add(A)`, and from there to `D._add(A)`. While fuctional, it is impossible to avoid redundancy `(C + D)._add(A + B)` will explore most of the same combinations - a problem that grows rapidly with the depth of the tree. By moving to a flat structure `A + B + C + D` it is trivial to find patterns with minimal effort. Moreover, most simplifications currently defined on `Add` and `Multiply` are already valid for multiple arguments, so the new structure does not add complexity.

### DIFF
--- a/nutils/evaluable.py
+++ b/nutils/evaluable.py
@@ -588,9 +588,11 @@ class Tuple(Evaluable):
 # The main evaluable. Closely mimics a numpy array.
 
 
-def add(a, b):
-    a, b = _numpy_align(a, b)
-    return Add(types.frozenmultiset((a, b)))
+def add(*args):
+    args = _numpy_align(*args)
+    if len(args) == 1:
+        return args[0]
+    return Add(types.frozenmultiset(args))
 
 
 def multiply(*args):
@@ -1774,41 +1776,34 @@ class Multiply(Array):
 class Add(Array):
 
     def __init__(self, funcs: types.frozenmultiset):
-        assert isinstance(funcs, types.frozenmultiset) and len(funcs) == 2, f'funcs={funcs!r}'
-        self.funcs = funcs
-        func1, func2 = funcs
-        assert equalshape(func1.shape, func2.shape) and func1.dtype == func2.dtype != bool, 'Add({}, {})'.format(func1, func2)
-        super().__init__(args=tuple(self.funcs), shape=func1.shape, dtype=func1.dtype)
+        assert isinstance(funcs, types.frozenmultiset), f'funcs={funcs!r}'
+        func, *others = funcs
+        assert all(equalshape(f.shape, func.shape) and f.dtype == func.dtype for f in others)
+        self.funcs = tuple(funcs)
+        super().__init__(args=tuple(funcs), shape=func.shape, dtype=func.dtype)
 
     @cached_property
     def _inflations(self):
-        func1, func2 = self.funcs
-        func2_inflations = dict(func2._inflations)
-        inflations = []
-        for axis, parts1 in func1._inflations:
-            if axis not in func2_inflations:
-                continue
-            parts2 = func2_inflations[axis]
-            dofmaps = set(parts1) | set(parts2)
-            if (len(parts1) < len(dofmaps) and len(parts2) < len(dofmaps)  # neither set is a subset of the other; total may be dense
-                    and self.shape[axis].isconstant and all(dofmap.isconstant for dofmap in dofmaps)):
-                mask = numpy.zeros(int(self.shape[axis]), dtype=bool)
-                for dofmap in dofmaps:
-                    mask[dofmap.eval()] = True
-                if mask.all():  # axis adds up to dense
-                    continue
-            inflations.append((axis, types.frozendict((dofmap, util.sum(parts[dofmap] for parts in (parts1, parts2) if dofmap in parts)) for dofmap in dofmaps)))
-        return tuple(inflations)
+        inflations_per_axis = [[] for i in range(self.ndim)]
+        for func in self.funcs:
+            for axis, parts in func._inflations:
+                inflations_per_axis[axis].append(parts)
+        return tuple((axis,
+            types.frozendict((dofmap, add(*(parts[dofmap] for parts in parts_per_func if dofmap in parts))) for dofmap in functools.reduce(set.union, map(set, parts_per_func))))
+                for axis, parts_per_func in enumerate(inflations_per_axis) if len(parts_per_func) == len(self.funcs))
 
     def _simplified(self):
-        func1, func2 = self.funcs
-        if func1 == func2:
-            return multiply(func1, 2)
-        for axes1 in func1._diagonals:
-            for axes2 in func2._diagonals:
-                if len(axes1 & axes2) >= 2:
-                    axes = sorted(axes1 & axes2)[:2]
-                    return diagonalize(takediag(func1, *axes) + takediag(func2, *axes), *axes)
+        for j, fj in enumerate(self.funcs):
+            if isinstance(fj, Add):
+                return add(*self.funcs[:j], *fj.funcs, *self.funcs[j+1:])
+            for i, fi in enumerate(self.funcs[:j]):
+                diags = [sorted(axesi & axesj)[:2] for axesi in fi._diagonals for axesj in fj._diagonals if len(axesi & axesj) >= 2]
+                unaligned1, unaligned2, where = unalign(fi, fj)
+                fij = diagonalize(takediag(fi, *diags[0]) + takediag(fj, *diags[0]), *diags[0]) if diags \
+                    else align(unaligned1 + unaligned2, where, self.shape) if len(where) != self.ndim \
+                    else fi._add(fj) or fj._add(fi)
+                if fij:
+                    return add(*self.funcs[:i], *self.funcs[i+1:j], *self.funcs[j+1:], fij)
         # NOTE: While it is tempting to use the _inflations attribute to push
         # additions through common inflations, doing so may result in infinite
         # recursion in case two or more axes are inflated. This mechanism is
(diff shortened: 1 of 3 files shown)


nutils/evaluable.py
```diff
@@ -1629,159 +1631,144 @@ class Multiply(Array):
...
+            if isinstance(fj, Multiply):
+                return multiply(*self.funcs[:j], *fj.funcs, *self.funcs[j+1:])
+            if fj._const_uniform == 1:
+                return multiply(*self.funcs[:j], *self.funcs[j+1:])
+            for i, parts in fj._inflations:
+                return util.sum(_inflate(multiply(f, *(_take(fi, dofmap, i) for fi in self.funcs[:j] + self.funcs[j+1:])), dofmap, self.shape[i], i) for dofmap, f in parts.items())
+            for axis1, axis2, *other in map(sorted, fj._diagonals):
+                return diagonalize(multiply(*(takediag(f, axis1, axis2) for f in self.funcs)), axis1, axis2)
+            for i, fi in enumerate(self.funcs[:j]):
+                unaligned1, unaligned2, where = unalign(fi, fj)
```
joostvanzwieten, line 1652: If we're outer-multiplying three arrays (e.g. `einsum('i,j,k->ijk', ...)`) then the result of this simplification depends on the order of the multiplication: the first two arguments are unaligned, multiplied and aligned, then multiplied with the third argument. This seems opposite to what the patch tries to achieve. Should this simplification not be part of stage two (`_optimized_for_numpy`)? We can anyhow unalign all arguments simultaneously.

nutils/evaluable.py
```diff
@@ -1629,159 +1631,144 @@ class Multiply(Array):
...
-        unaligned2, where2 = unalign(func2)
-        return Einsum((unaligned1, unaligned2), (where1, where2), tuple(range(self.ndim)))
+        for j, fj in enumerate(self.funcs):
+            if fj._const_uniform == -1:
+                return Negative(multiply(*self.funcs[:j], *self.funcs[j+1:]))
+            if isinstance(fj, Sign) and fj.func in self.funcs:
+                fi = fj.func
+                i = self.funcs.index(fi)
+                a, b = sorted([i,j])
+                return multiply(*self.funcs[:a], *self.funcs[a+1:b], *self.funcs[b+1:], Absolute(fi))
```
joostvanzwieten, line 1666: Nice! Why here and not in `_simplified`? Is `fj` never encapsulated in a `Transpose`?

nutils/evaluable.py
```diff
@@ -1774,41 +1776,36 @@ def _intbounds_impl(self):
...
-        return tuple(inflations)
+        inflations_per_axis = [[] for i in range(self.ndim)]
+        for func in self.funcs:
+            for axis, parts in func._inflations:
+                inflations_per_axis[axis].append(parts)
+        return tuple((axis,
+            types.frozendict((dofmap, add(*(parts[dofmap] for parts in parts_per_func if dofmap in parts)))
+                for dofmap in functools.reduce(set.union, map(set, parts_per_func))))
+                    for axis, parts_per_func in enumerate(inflations_per_axis)
+                        if len(parts_per_func) == len(self.funcs))
```
joostvanzwieten, line 1795: The indentation really confuses me. The last two lines belong to the generator starting at line 1791, but the line before, with less indentation, belongs to the inner generator starting at line 1791. Also it is hard to see where the tuple `(axis, ...)`, the item of the outer generator, ends. Can you please reconsider this?

nutils/evaluable.py
```diff
@@ -1774,41 +1776,36 @@ def _intbounds_impl(self):
...
-        for axes1 in func1._diagonals:
-            for axes2 in func2._diagonals:
-                if len(axes1 & axes2) >= 2:
-                    axes = sorted(axes1 & axes2)[:2]
-                    return diagonalize(takediag(func1, *axes) + takediag(func2, *axes), *axes)
+        for j, fj in enumerate(self.funcs):
+            if isinstance(fj, Add):
+                return add(*self.funcs[:j], *fj.funcs, *self.funcs[j+1:])
+            for i, fi in enumerate(self.funcs[:j]):
+                diags = [sorted(axesi & axesj)[:2] for axesi in fi._diagonals for axesj in fj._diagonals if len(axesi & axesj) >= 2]
```
joostvanzwieten, line 1802: Swapping `Diagonalize` and `Add` for two arguments suffers from the same ordering problem as `unalign` for `Multiply`. Do we actually need this simplification (for more than two arguments) before `_optimize_for_numpy`?

nutils/evaluable.py
gertjanvanzwieten: The `unalign` method identifies insertions that are common to the arguments, so this is not a `i,j,k` situation but rather the multiplication of two `InsertAxes` objects. Which means that this could probably equally reside in `InsertAxis._multiply` -- still not sure which is better. Note that the old `Multiply._simplified` featured the same simplification.

nutils/evaluable.py
gertjanvanzwieten: Operations like `negative` and `abs` are implemented using basic building blocks to facilitate simplifications, and only during optimization replaced by their `Pointwise` counterparts. This was already the case and this PR is not aiming to change this. The `Transpose` issue is relevant (though rare) but not easily solvable, and the previous implementation suffered the same problem.

nutils/evaluable.py
gertjanvanzwieten: Promoting `Diagonalize` reduces flops, and as such is a relevant modification for the simplification step. It also exposes structure to follow-up operations that can propagate the structure further. The optimization step, in my view, is a more local sweep that combines and/or replaces operations for more efficient evaluation. Note also that this swap was always part of simplify; I don't believe this PR introduces any new simplifications or optimizations.

nutils/evaluable.py
joostvanzwieten: I understand that this gives the same simplification as before, and that is a valid point to keep it like this.

To clarify, my question concerns the following situation:

```python
from nutils import evaluable
arg = lambda name, length: evaluable.Argument(name, (evaluable.asarray(length),))
abc = evaluable.einsum('i,j,k->ijk', arg('a', 2), arg('b', 3), arg('c', 4)).simplified
```

With this PR (and current master) the tree of `abc` (note: simplified!) is

```
%0 = Multiply; f:2,3,4
: %1 = InsertAxis; f:2,3,4
| : %2 = Multiply; f:2,3
| | : %3 = InsertAxis; f:2,3
| | | : Argument; a; f:2
| | | : 3
| | : %4 = Transpose; 1,0; f:2,3
| |   : %5 = InsertAxis; f:3,2
| |     : Argument; b; f:3
| |     : 2
| : 4
: %6 = Transpose; 1,2,0; f:2,3,4
  : %7 = InsertAxis; f:4,2,3
    : %8 = InsertAxis; f:4,2
    | : Argument; c; f:4
    | : 2
    : 3
```

The choice to group and unalign the first two items of the multiplication seems arbitrary to me. Why not the first and the third. Or the second and the third. I have no objection against applying unalign to all arguments at once.

For reference, this is the tree of `abc` without unalign simplification:

```
%0 = Multiply; f:2,3,4
: %1 = InsertAxis; f:2,3,4
| : %2 = InsertAxis; f:2,3
| | : Argument; a; f:2
| | : 3
| : 4
: %3 = Transpose; 1,0,2; f:2,3,4
| : %4 = InsertAxis; f:3,2,4
|   : %5 = InsertAxis; f:3,2
|   | : Argument; b; f:3
|   | : 2
|   : 4
: %6 = Transpose; 1,2,0; f:2,3,4
  : %7 = InsertAxis; f:4,2,3
    : %8 = InsertAxis; f:4,2
    | : Argument; c; f:4
    | : 2
    : 3
```

nutils/evaluable.py
joostvanzwieten: Perhaps solvable by a property `_unsigned` such that `Sign(func._unsigned)` is equivalent to `func` if not `None` (similar to `_unalign`, `_inflations`, `_diagonals`), but let's leave that for the future.

nutils/evaluable.py
gertjanvanzwieten: In terms of flops the former is better though, even if the choice is arbitrary. The fact that simplify does not bring the function in a canonical form is unfortunate but not specific to this operation. To mention just one example: the two `InsertAxis` operations follows by a `Transpose` could be arbitrarily swapped. The same applies to pretty much any sequence of operations of the same type. Not to say that the problem is not worth exploring, but I do not think the present PR is the appropriate opportunity. As for its intentions I have added some context in the description.

gertjanvanzwieten: But simplify is the flops reducing step; optimize merely brings it in a form that is more efficient to evaluate in a given backend (just numpy for now). This is why take is pushed up and inflate pulled down during simplify, and in this case inserted axes are merged where possible. I think it's important to stay close to what each step is aiming to accomplish and not blur any lines, given the complexity of the framework. Ultimately we will want to make simplify more powerful than the current greedy algorithm and then we will benefit from maintaining a separation of tasks.

nutils/evaluable.py
gertjanvanzwieten: I reformatted it to make more sense but at the expense of a very long line.

nutils/evaluable.py
joostvanzwieten: Consider this example: `a`, `b` and `c` are 3D arrays, each with two inserted axes (think of meshgrid). With this PR simplifying `a + a * b * c` gives `a + (a * b) * c` and simplifying `a + c * b * a` gives `a * (b * c + 1)`  (transposes and inserted axes omitted). The latter is made possible by the PR. Example code:

```python
from nutils.evaluable import align, asarray, Argument
shape = tuple(map(asarray, (2, 3, 4)))
a = align(Argument('a', shape[0:1]), (0,), shape)
b = align(Argument('b', shape[1:2]), (1,), shape)
c = align(Argument('c', shape[2:3]), (2,), shape)
f1 = a + a * b * c
f2 = a + c * b * a
print(f1.simplified.asciitree())
print(f2.simplified.asciitree())
```

I don't know which of the two is better and there will always be situations where the order of the operations affect the simplification, but this one is a avoidable. For instance, we could add an `_unmultiply` property that undoes the unalign simplification.

As I see it the objective of this PR is twofold:

* Enable more simplifications by providing flattened adds and multiplies.
* More efficient `{Add,Multiply}.evalf` by not having to allocate intermediate arrays.

As it stands now they are in an arms race.

gertjanvanzwieten:
Closing in favour of #818

PR #789, opened 2023-05-09
exception to non-scalar passsed to solver.optimizer 

Sbhat92:


nutils/solver.py
```diff
@@ -763,6 +763,8 @@ def _optimize(target, functional: evaluable.asarray, constrain, arguments, tol:
     dtype = _determine_dtype(target, (functional,), lhs0, constrain)
     mask, vmask = _invert(constrain, target)
     lhs, vlhs = _redict(lhs0, target, dtype)
+    if functional.shape != ():
```
gertjanvanzwieten, line 766: This is valid, but a slightly cleaner check is `functional.ndim != 0`.

nutils/solver.py
```diff
@@ -763,6 +763,8 @@ def _optimize(target, functional: evaluable.asarray, constrain, arguments, tol:
...
+        raise ValueError('The objective function must be a scalar function of a single variable')
```
gertjanvanzwieten, line 767: Optimize can be used for multiple variables, so the second part of this message is false. A valid message could be simply "the objective function must be scalar valued".

gertjanvanzwieten (changes requested):

Sbhat92:
Can I add a test case as well?

gertjanvanzwieten:
> Can I add a test case as well?

Absolutely! I think the best place for that would be in tests/test_solver.py in the `optimize` class.

Issue #791, opened 2023-05-14
incorrect weights in zipped sample


joostvanzwieten:
Let sample `Z` be the zip of two samples `X` and `Y`. It is guaranteed that the order in which `Z` traverses the elements of `X` is monotonic. Not strict monotonic, however: if the points `p` of some element `i` of `X` are divided over `n >= 2` elements of `Y`, then element `i` is evaluated `n` times, each with a different subset `s_j` of points `p` and weights `w`. If the subsets `(s_j)_j` are consecutive slices, `max(s_j) < min(s_{j+1})`, then everything works fine. If not, for the `j`-th evaluation of element `i` points `p[s_j]` is used together with weights `w[t_j]` instead of `w[s_j]`, where `(t_j)_j` is a sequence of consecutive slices such that the sizes of the subsets `s_j` and `t_j` are equal.

Minimal example that triggers the bug:

```python
X, x = mesh.line([1, 5], space='X')
Y, y = mesh.line([0, 3, 4], space='Y')
smpl = X.sample('gauss', 6)
zipped = smpl.zip(Y.locate(y, smpl.eval(x) % 4, tol=1e-10))
zipped.integrate(x * function.J(x)) # returns != 12
```

The third point of `smpl` lies in the second element of `Y`, the other three in the first element of `Y`.

Affected versions:

- [ ] master (#792)
- [ ] v7

PR #792, opened 2023-05-14
fix weights of Sample.zip

joostvanzwieten:
Let sample `Z` be the zip of two samples `X` and `Y`. It is guaranteed that the order in which `Z` traverses the elements of `X` is monotonic. Not strict monotonic, however: if the points `p` of some element `i` of `X` are divided over `n >= 2` elements of `Y`, then element `i` is evaluated `n` times, each with a different subset `s_j` of points `p` and weights `w`. If the subsets `(s_j)_j` are consecutive slices, `max(s_j) < min(s_{j+1})`, then everything works fine. If not, for the `j`-th evaluation of element `i` points `p[s_j]` is used together with weights `w[t_j]` instead of `w[s_j]`, where `(t_j)_j` is a sequence of consecutive slices such that the sizes of the subsets `s_j` and `t_j` are equal. This patch fixes this problem and adds a test for this situation.

Fixes: #791

### DIFF
--- a/nutils/sample.py
+++ b/nutils/sample.py
@@ -734,8 +734,10 @@ class _Zip(Sample):
         return evaluable.Take(evaluable.Constant(self._indices), self._getslice(ielem))
 
     def get_evaluable_weights(self, ielem):
-        weights = self._samples[0].get_evaluable_weights(evaluable.Take(evaluable.Constant(self._ielems[0]), ielem))
-        return evaluable.Take(weights, self._getslice(ielem))
+        ielem0 = evaluable.Take(evaluable.Constant(self._ielems[0]), ielem)
+        slice0 = evaluable.Take(evaluable.Constant(self._ilocals[0]), self._getslice(ielem))
+        weights = self._samples[0].get_evaluable_weights(ielem0)
+        return evaluable._take(weights, slice0, axis=0)
 
 
 class _TakeElements(_TensorialSample):
--- a/tests/test_sample.py
+++ b/tests/test_sample.py
@@ -300,6 +300,21 @@ Zip(etype='square')
 Zip(etype='triangle')
 Zip(etype='mixed')
 
+class ZipCornerCases(TestCase):
+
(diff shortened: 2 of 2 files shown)


gertjanvanzwieten:
I'm a bit confused by the description of this PR: "When the order of the points of the first sample in a zip is not maintained" -- isn't the core principle of zip that it maintains the points order of all samples? Don't you mean the order of _elements_?

Also, was this just an ordering issue or could `get_evaluable_weights` fail? Looking at the old implementation I'm surprised it worked at all, as `_getslice` provides indices in a global array whereas `weights` are pre-element. I have to admit I'm having a hard time tracing through the mountain of renumberings so if you have some strategic comments you can leave behind now that you have things fresh in your head that would be valuable.

joostvanzwieten:
> I'm a bit confused by the description of this PR: "When the order of the points of the first sample in a zip is not maintained" -- isn't the core principle of zip that it maintains the points order of all samples? Don't you mean the order of _elements_?

Yes, the order of points in the zip is the same as the order in the first sample of the zip. However, the order of evaluation might differ. The bug is triggered when it differs.
 
> Also, was this just an ordering issue or could `get_evaluable_weights` fail? Looking at the old implementation I'm surprised it worked at all, as `_getslice` provides indices in a global array whereas `weights` are pre-element. I have to admit I'm having a hard time tracing through the mountain of renumberings so if you have some strategic comments you can leave behind now that you have things fresh in your head that would be valuable.

I've updated the commit and PR message. Hopefully that clarifies some of the confusion.

gertjanvanzwieten:
The new description is is very helpful, I get the issue now. Could you rebase to the master? (I pulled #790 in the meantime).

gertjanvanzwieten (approved):

joostvanzwieten:
> The new description is is very helpful, I get the issue now. Could you rebase to the master? (I pulled #790 in the meantime).

Rebased.

gertjanvanzwieten (approved):

PR #797, opened 2023-05-30
extend support for nested _mul.tri

gertjanvanzwieten:
For a multiplied sample A * B, the .tri implementation added trivial implementations for the situation that either A or B consisted of only one point, or A is one-dimensional. This PR extends this logic to any multiplication chain A0 * A1 * ... with support for all left and right multiplications of single-point or one-dimensional samples. In particular this extends support to 3D structured boundaries.

### DIFF
--- a/nutils/sample.py
+++ b/nutils/sample.py
@@ -15,7 +15,7 @@ selected sample points, and is typically used in combination with the "bezier"
 set.
 '''
 
-from . import types, points, _util as util, function, evaluable, parallel, numeric, matrix, sparse, warnings
+from . import types, points, _util as util, function, evaluable, parallel, matrix, sparse, warnings
 from .pointsseq import PointsSequence
 from .transformseq import Transforms
 from ._backports import cached_property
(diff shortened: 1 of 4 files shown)


nutils/sample.py
```diff
@@ -595,6 +595,103 @@ def __call__(self, func: function.IntoArray) -> function.Array:
...
+    elif ndims1 == 1: # Left multiplication by a 1D sample
+        # Multiply the left 1D tri by the right tri and triangulate using a
+        # simplex strip.
+        tri = _simplex_strip(tri_outer).reshape(-1, ndims2+2)
+    elif ndims2 == 1: # Right multiplication by a 1D sample
+        # Multiply the left 1D tri by the right tri and triangulate using a
+        # simplex strip.
+        tri = _simplex_strip(tri_outer.swapaxes(-2,-1)).reshape(-1, ndims1+2)
+    else:
+        raise NotImplementedError(f'tri not supported for {ndims1}D x {ndims2}D multiplication')
```
joostvanzwieten, line 642: What do you think of swapping `tri1` and `tri2` after singleton expansion of the first axis if `tri1.shape[-1] > tri2.shape[-1]` and dropping the second and the fourth branch? Same for the hull.

joostvanzwieten (approved):

nutils/sample.py
gertjanvanzwieten: Done - and we can even do it before singleton expansion of the first axis as we don't make any promises about the order of the returned triangles.

joostvanzwieten (approved):

PR #859, opened 2024-03-01
Evaluable to Python code compiler

joostvanzwieten:
Previously `Evaluable`s were evaluated by looping over all dependencies and calling `Evaluable.evalf`. For `1 + 2 + 3` (not simplified) this boils down to

```python
serialized = [
    [Constant(1).evalf],
    [Constant(2).evalf],
    [Constant(3).evalf],
    [Add.evalf, 0, 1],
    [Add.evalf, 3, 2],
]
def eval():
    values = []
    for op, *args in serialized:
        values.append(op(*map(values.get, args)))
    return values[-1]
```

This PR replaces this mechanism by generating Python code via the new `evaluable.compile()` function, which produces:

```python
c0 = 1
c1 = 2
c2 = 3
def eval():
    v3 = numpy.add(c0, c1)
    v4 = numpy.add(v3, c2)
    return v4
```

In addition, subtrees that are constant are cached for a second call, providing a slight performance improvement.


nutils/evaluable.py
```diff
@@ -4437,6 +4437,10 @@ def _node(self, cache, subgraph, times, unique_loop_ids):
         cache[self] = node = self._node_loop_body(loopcache, loopgraph, looptimes, unique_loop_ids)
         return node
 
+    @cached_property
+    def arguments(self):
+        return super().arguments - frozenset({self.index})
+
```
gertjanvanzwieten, line 4128: The commit message explains that the `Loop.arguments` serves to shadow the default implementation, yet it calls `super().arguments`. What am I missing?

nutils/evaluable.py
```diff
@@ -5708,7 +5333,7 @@ def _compile_to_pyast(funcs, globals, evaluables, new_index, cache, stats, paral
             ndependents[func] = 1
     while stack:
         func = stack.pop()
-        for arg in ((func.init_arg, func.body_arg) if isinstance(func, Loop) else func._Evaluable__args):
+        for arg in func._Evaluable__args:
```
gertjanvanzwieten, line 5337: Is it correct that the iterable gains the loop's length here? (honestly I'm more surprised by the fact that it wasn't there before, but apparently that worked)

nutils/evaluable.py
```diff
@@ -5029,6 +5029,7 @@ def eval_sparse(funcs: AsEvaluableArray, **arguments: typing.Mapping[str, numpy.
         yield data
 
 
+@functools.lru_cache(256)
```
gertjanvanzwieten, line 5032: Have you considered a weakref-like module level dictionary, with keys formed of `tuple(ref(func, removeme) for func in funcs)` plus whatever else is relevant, and `removeme` a callback that removes the entry? That way it is triggered as soon as any involved function is garbage collected.

nutils/evaluable.py
```diff
@@ -5243,6 +5321,556 @@ def eval_sparse(funcs: AsEvaluableArray, **arguments: typing.Mapping[str, numpy.
...
+    ret_fmt = []
+    funcs = []
+    MakeTuple = collections.namedtuple('MakeTuple', ('n'))
+    while stack:
+        obj = stack.pop(0)
+        if isinstance(obj, MakeTuple):
+            m = len(ret_fmt) - obj.n
+            ret_fmt[m:] = ['(' + ', '.join(ret_fmt[m:]) + (',)' if obj.n == 1 else ')')]
+        elif isinstance(obj, (tuple, list)):
+            stack[:0] = [*obj, MakeTuple(len(obj))]
```
gertjanvanzwieten, line 5129: Wouldn't it be more natural to use a right sided stack, so that you can use `append` and `extend`?

gertjanvanzwieten (commented):
A last point relating to c30f1565: if the issue is only with nested loops, did you consider making the ids unique by simply prepending the name of the enclosing loop? That way you can stick to strings and it might even have a slight debugging advantage. Plus you could maybe even make the change on the spot, like in the Loop constructor, so that you can always assume uniqueness and even remove the flag in 47f7e1b7.

joostvanzwieten:
> I created commit [8159737](https://github.com/evalf/nutils/commit/8159737c06cd882e515d0211be8a7904dd9d9f00) as a suggestion for making comments inline, simply by making them an argument of `Assign`. It's crude but effective, and I personally find the resulting code easier to follow.

I've replaced `_pyast.Comment` with `_pyast.CommentBlock(comment, statements)`, which appends `comment` to the next line if the `statements` is a single line.

nutils/evaluable.py
joostvanzwieten: Fixed.

nutils/evaluable.py
joostvanzwieten: Fixed.

nutils/evaluable.py
joostvanzwieten: The `lru_cache` is a temporary solution until there's a solver 2.0 that avoids repeated calls to `compile`. I've reduced the size of the cache to 32.

nutils/evaluable.py
joostvanzwieten: Fixed by adding `length` to commit 'add inplace compilation'.

joostvanzwieten: Fixed in the referenced commit. Note that in commit 'remove unused Evaluable.evalfs,serialized etc.' the eval args are equal to `init_arg, body_arg`, so we can again use `super().arguments`.

nutils/_util.py
```diff
@@ -1082,4 +1082,39 @@ def shallow_replace(func, *funcargs, **funckwargs):
...
+    for root in roots:
+        if id(root) not in seen:
+            queue.append(root)
+            seen.add(id(root))
+    while queue:
+        node = queue.pop()
+        for dep in visit_node(node) or ():
+            if id(dep) not in seen:
+                queue.append(dep)
+                seen.add(id(dep))
```
gertjanvanzwieten, line 1117: Doesn't this have a simpler implementation in
```python
stack = IDSet(roots)
while stack:
    stack.update(visit_node(stack.pop()) or ())
```
This gets even cleaner if the visitor returns an empty list to terminate instead of `None`, which also opens it up to generators.
(Note I renamed queue to stack since we're appending and popping from the same side)

gertjanvanzwieten (approved):
Just this one thought, otherwise go ahead and merge!

nutils/_util.py
joostvanzwieten: Every node should be visited only once. That's where `seen` is for.

gertjanvanzwieten: Right, didn't think of the items that are popped and forgotten.

gertjanvanzwieten (approved):

PR #861, opened 2024-03-08
rewrite evaluable loops using Loop base class

joostvanzwieten:
This PR merges the common parts of the evaluable loops into a base class.


nutils/evaluable.py
```diff
@@ -327,50 +328,75 @@ def _deep_flatten_constants(self):
...
+            #         ...
+            #
+            # which is not in the set of candidates and we'd miss the
+            # opportunity to combine B' with F. Or worse: we combine B with F,
+            # ignore the former (`ArrayFromTuple`) *and* keep B'.
+            loops_sans_deps = candidates.copy()
+            for loop in candidates:
+                loops_sans_deps.difference_update(loop._loop_deps - util.IDSet([loop]))
+            index = next(iter(loops_sans_deps)).index
+            loops = util.IDSet([loop for loop in loops_sans_deps if loop.index == index])
```
gertjanvanzwieten, line 381: I think this is faster (no set operations, just inclusion checks):
```
loops = []
for loop in candidates:
    if loops and loop.index != loops[0].index:
        continue # take one index at a time
    if all(loop not in other._loop_deps for other in candidates if other is not loop):
        loops.append(loop)
```

nutils/evaluable.py
```diff
@@ -327,50 +328,75 @@ def _deep_flatten_constants(self):
...
+            assert loops
+
+            if len(loops) >= 2:
+                replacements = util.IDDict()
+                combined = _LoopTuple(tuple(loops), index.name, index.length)
+                combined = combined._combine_loops(combined.body_arg._loop_deps - combined._loop_deps)
+                for i, loop in enumerate(loops):
+                    kwargs = {}
+                    replacements[loop] = ArrayFromTuple(combined, i, loop.shape, loop.dtype)
+                self = util.shallow_replace(replacements.get, self)
```
gertjanvanzwieten, line 392: Can't we wait till the end to apply all replacements at once?

gertjanvanzwieten (changes requested):

nutils/evaluable.py
joostvanzwieten: Done.

nutils/evaluable.py
joostvanzwieten: No, I don't think so. `shallow_replace` stops at the shallowest replacement, so we'd miss replacements that are a dependency of the shallowest replacement.

PR #874, opened 2024-05-24
replace coverage with custom framework

joostvanzwieten:
Python 3.12 has a significant regression in coverage analysis. Related issue: https://github.com/python/cpython/issues/107674 . This PR replaces coverage with a custom framework built using `sys.monitoring`, which is only available as of Python 3.12.


devtools/gha/unittest.py
```diff
@@ -0,0 +1,52 @@
...
+import json
+import os
+from pathlib import Path
+import sys
+import unittest
+
+source = importlib.util.find_spec('nutils').origin
+assert source.endswith(os.sep + '__init__.py')
+source = source[:-11]
+coverage = {}
```
gertjanvanzwieten, line 12: A comment explaining the values 0, 1 and 2 would be useful here.

devtools/gha/unittest.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def start(code, _):
+        if isinstance(code.co_filename, str) and code.co_filename.startswith(source) and not sys.monitoring.get_local_events(sys.monitoring.COVERAGE_ID, code):
+            if (file_coverage := coverage.get(code.co_filename)) is None:
+                with open(code.co_filename, 'rb') as f:
+                    nlines = sum(1 for _ in f)
+                coverage[code.co_filename] = file_coverage = [0] * (nlines + 1)
+            for _, _, l in code.co_lines():
+                if l:
+                    file_coverage[l] = 1
```
gertjanvanzwieten, line 24: Does this not imply that functions that are never visited do not count as misses?

nutils/SI.py
```diff
@@ -15,7 +15,7 @@
     >>> from nutils import SI
     >>> v = SI.parse('7μN*5h/6g')
 
-The Quantity constructor recognizes the multiplication (\*) and division (/)
+The Quantity constructor recognizes the multiplication (\\*) and division (/)
```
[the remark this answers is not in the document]
joostvanzwieten: This is to prevent interpreting the asterisk as the start of a span of italic text. See https://docutils.sourceforge.io/docs/user/rst/quickstart.html#toc-entry-2 .

devtools/gha/unittest.py
joostvanzwieten: No, because we call `start` recursively for all constants in the code object, which includes functions defined in the code object.

devtools/gha/unittest.py
joostvanzwieten: comment added

devtools/gha/unittest.py
gertjanvanzwieten: Neat.

PR #876, opened 2024-06-11
Dataclass

gertjanvanzwieten:
This PR changes the base class of `evaluable.Evaluable` to `types.DataClass`, which is a newly introduced base class akin to Python's built in dataclass but implemented as a singleton.

### DIFF
--- a/nutils/evaluable.py
+++ b/nutils/evaluable.py
@@ -140,7 +140,7 @@ class Evaluable(types.Singleton):
     def __init__(self, args: typing.Tuple['Evaluable', ...]):
         assert isinstance(args, tuple) and all(isinstance(arg, Evaluable) for arg in args), f'args={args!r}'
         super().__init__()
-        self.__args = args
+        self.dependencies = args
 
     @staticmethod
     def evalf(*args):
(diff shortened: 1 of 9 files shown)


nutils/evaluable.py
```diff
@@ -134,6 +134,10 @@ class ExpensiveEvaluationWarning(warnings.NutilsInefficiencyWarning):
     pass
 
 
+def _abstract_property(self):
+    raise NotImplementedError(f'{self.__class__.__name__} fails to implement dependencies')
```
joostvanzwieten, line 138: The error message refers to some property `dependencies`, but `_abstract_property` is used from `shape` and `dtype`.

nutils/types.py
```diff
@@ -290,6 +290,78 @@ class Singleton(Immutable, metaclass=SingletonMeta):
...
+            self.__post_init__()
+            cls.__cache[bound.args] = self
+        return self
+
+
+class DataClass(metaclass=DataClassMeta):
+    '''
+    Base class for dataclass-like types, for which an assignment only
+    constructor is inferred from type annotations. Post processing is possible
+    via an optional __post_init__ method. The class adds equality tests,
```
joostvanzwieten, line 320: ``` :meth:`__post_init__` ```

gertjanvanzwieten: fixed

nutils/evaluable.py
gertjanvanzwieten: I added an `abstract_property` descriptor class in `_util` that gets hold of the attribute name via `__set_name__`.

nutils/types.py
gertjanvanzwieten: correction: sphinx didn't like the change so I changed it to double backticks instead

joostvanzwieten (approved):

PR #884, opened 2024-09-25
Introducing solver.System

gertjanvanzwieten:


nutils/matrix/_scipy.py
```diff
@@ -82,7 +82,8 @@ def mycallback(arg):
                 if callback:
                     callback(res)
                 reformat(100 * numpy.log10(max(atol, res)) / numpy.log10(atol))
-            lhs, status = solverfun(self.core, rhs, M=precon, tol=0., atol=atol, callback=mycallback, **solverargs)
+            solverargs['rtol' if scipy.version.version >= '1.12' else 'tol'] = 0
```
joostvanzwieten, line 85: It looks like the documented interface for SciPy's version is `scipy.__version__`. See https://github.com/scipy/scipy/blob/main/scipy/__init__.py#L39 . There's no mention of `scipy.version`.

gertjanvanzwieten: Fixed

nutils/numeric.py
```diff
@@ -676,4 +676,20 @@ def sanitize_einsum_subscripts(subscripts, *shapes):
     return (*in_, out)
 
 
+def compress_indices(indices, length):
```
joostvanzwieten, line 679: Suggestion: It took me a bit of time to understand that the returned array is an array of insertion positions of `range(length+1)` in `indices`. What about

```python
'''Returns the insertion positions of ``range(length+1)`` in the monotonically increasing ``indices``.'''
```

nutils/numeric.py
```diff
@@ -676,4 +676,20 @@ def sanitize_einsum_subscripts(subscripts, *shapes):
...
+    '''Compress an array of monotonically increasing indices.
+
+    The compressed array `c` is such that indices[c[i]:c[i+1]] == i for any
+    value of i in range(length).'''
+
+    step = numpy.empty(len(indices)+1, dtype=int)
```
joostvanzwieten, line 685: Suggestion: You can save a store to (line 688) and load from `step` (line 689) by using `numpy.subtract(indices[1:], indices[:-1], out=step[1:-1])` as follows:

```python
def compress_indices(indices, length):
    if not len(indices):
         return numpy.zeros((length+1,), int)
     step = numpy.empty(len(indices)+1, dtype=int)
     step[-1] = length - indices[-1]
     numpy.subtract(indices[1:], indices[:-1], out=step[1:-1])
     step[0] = indices[0] + 1
     nz, = step.nonzero()
     return numpy.repeat(nz, step[nz]) # NOTE fails if any step is negative
```

tests/test_numeric.py
```diff
@@ -415,3 +415,18 @@ def test_matrix_2dindex(self):
...
         self.assertEqual(A.tolist(), [[1, 0, 0, 2], [3, 0, 0, 4]])
+
+
+class compress_indices(TestCase):
+
+    def test(self):
+        indices = numpy.array([0, 2, 2, 2, 5, 6, 7])
+        length = 8
+        c = numeric.compress_indices(indices, length)
+        for n in range(length):
```
joostvanzwieten, line 426: Why not just compute the desired value by hand? This unittest doesn't test the length of `c` (a `c` that is too long goes unnoticed).

nutils/matrix/__init__.py
```diff
@@ -23,27 +23,87 @@ def backend(matrix: str = 'auto'):
...
+    if not (rowptr.ndim == 1 and
+            rowptr.dtype.kind in 'ui' and
+            rowptr[0] == 0 and
+            all(rowptr[1:] >= rowptr[:-1]) and
+            rowptr[-1] == len(values)):
+        raise MatrixError('assemble received invalid row indices')
+    if not (colidx.ndim == 1 and
+            colidx.dtype.kind in 'ui' and
+            len(colidx) == len(values) and
+            all(all(colidx[i+1:j] >= colidx[i:j-1]) for i, j in zip(rowptr, rowptr[1:]) if i != j) and
```
joostvanzwieten, line 63: I haven't benchmarked this line, but have you considered something along these lines (unchecked)?

```python
colidx_is_increasing = numpy.empty((len(colidx)+1,), bool)
numpy.greater_equal(colidx[1:], colidx[:-1], out=colidx_is_increasing[1:])
colidx_is_increasing[rowptr] = True
if not (colidx_is_increasing.all()):
    raise MatrixError('assemble received invalid column indices')
```

joostvanzwieten (commented):
Stopped at 'Add matrix.assemble_block_csr`. Continuing review tomorrow.

nutils/numeric.py
gertjanvanzwieten: Changed it to *Return insertion positions in monotonically increasing index vector* for brevity, with the following elaboration:

The compressed array ``c`` is such that ``indices[c[i]:c[i+1]] == i`` for any integer ``0 <= i <= length``. Fails with ``ValueError`` if ``indices`` are out of bounds or not monotonically increasing; otherwise fully equivalent to ``indices.searchsorted(numpy.arange(length + 1))``.

nutils/numeric.py
gertjanvanzwieten: Modified, plus added out of bounds check.

tests/test_numeric.py
gertjanvanzwieten: I think unit tests should check for properties of the return value where possible, rather than compare against precomputed values that cannot be easily verified. I added the length check though.

nutils/matrix/__init__.py
gertjanvanzwieten: I considered that, but favoured the conciseness of the inline test, which also has fewer comparisons. I get that the proposed form benefits from fewer instructions but doubt the difference is large enough to merit the added code complexity.

joostvanzwieten: Benchmark of the first call to `assemble_csr` in script `example/finitestrain.py`: 554 μs vs 62.1 μs. Zooming in on the specific test: 498 μs vs 8.25 μs.

gertjanvanzwieten: Ok fine :)

nutils/solver.py
```diff
@@ -180,6 +181,271 @@ def __call__(self, res0, dres0, res1, dres1):
...
+        self.__argsizes = {t: numpy.prod(shape, dtype=int) for t, shape in self.argshapes.items()}
+        self.__eval = evaluable.compile((tuple(tuple(map(evaluable.as_csr, row)) for row in block_matrix), tuple(block_vector), value))
+        self.is_linear = not any(arg.name in trials for row in block_matrix for col in row for arg in col.arguments)
+        self.is_constant = self.is_linear and not any(col.arguments for row in block_matrix for col in row)
+        self.__matrix_cache = None
+
+    @functools.cached_property
+    def __nutils_hash__(self):
+        from linecache import cache # we abuse the line cache to hash self.eval
+        return types.nutils_hash((self.trials, ''.join(cache[self.__eval.__code__.co_filename][2])))
```
joostvanzwieten, line 214: I think we can add a hash to the return value of `evaluable.compile`. Then we don't have to fetch the code from the line cache, because the code is generated there.

joostvanzwieten (approved):

PR #885, opened 2024-10-15
Introducing function.factor

gertjanvanzwieten:


nutils/evaluable.py
```diff
@@ -217,6 +222,24 @@ def _compile_expression(self, py_self: _pyast.Variable, *args: _pyast.Expression
 
         return py_self.get_attr('evalf').call(*args)
 
+    def argument_degree(self, argument):
+        '''return the heighest power of argument of self is polynomial,
```
joostvanzwieten, line 226: heighest -> highest

nutils/evaluable.py
```diff
@@ -217,6 +222,24 @@ def _compile_expression(self, py_self: _pyast.Variable, *args: _pyast.Expression
...
+        or raise NotPolynomal otherwise.'''
+        # IMPORTANT: since we are tracking only the highest power, we cannot
+        # ever lower a power, e.g. via division. To see this, consider the sum
+        # of a 0th and 1st power, registered as degree 1. This would be lowered
+        # via division to 0, thereby suggesting that the evaluable is constant
+        # while in reality it is not polynomial.
+        if argument not in self.arguments:
```
joostvanzwieten, line 233: I don't think this is an actual problem, but the arguments of the shape of `ArrayFromTuple` are not included in the `ArrayFromTuple.arguments`, which renders this test invalid if `argument` is part of the shape, but not part of the tuple. The proper fix is to include the shape args in `ArrayFromTuple.arguments`.

nutils/evaluable.py
```diff
@@ -4986,6 +5051,9 @@ def _assparse(self):
                 chunks.append(tuple(loop_concatenate(arr, self.index) for arr in (*elem_indices, elem_values)))
         return tuple(chunks)
 
+    def _argument_degree(self, argument):
+        return self.func._argument_degree(argument)
```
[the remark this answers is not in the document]
gertjanvanzwieten: fixed

nutils/evaluable.py
gertjanvanzwieten: fixed

nutils/evaluable.py
gertjanvanzwieten: The ArrayFromTuple shape can only depend on an argument if the evaluable itself also depends on that argument, otherwise it could not evaluate to an array of the correct shape, so the default implementation should suffice.

joostvanzwieten (approved):

PR #887, opened 2024-10-21
Improvements for solver.System

gertjanvanzwieten:


nutils/solver.py
```diff
@@ -436,11 +436,13 @@ def optimize(self, *, droptol: Optional[float], arguments: Dict[str, numpy.ndarr
...
         jac, res, val = self.assemble(arguments)
-        nosupp = ~jac.rowsupp(droptol)
-        dx = -jac.solve(res, constrain=iscons | nosupp, **_copy_with_defaults(linargs, symmetric=self.is_symmetric))
+        mycons = iscons | ~jac.rowsupp(droptol)
+        dx = -jac.solve(res, constrain=mycons, **_copy_with_defaults(linargs, symmetric=self.is_symmetric))
         log.info(f'optimal value: {val+.5*(res@dx):.1e}') # val(x + dx) = val(x) + res(x) dx + .5 dx jac dx
         x += dx
-        x[nosupp & ~iscons] = numpy.nan
+        for trial, i, j in zip(self.trials, self.__trial_offsets, self.__trial_offsets[1:]):
+            log.info(f'constrained {(~mycons[i:j]).sum()} degrees of freedom of u')
```
joostvanzwieten, line 443: Instead of `u` the name of the trial argument should be formatted.

Perhaps we also want to see the total amount of dofs `j - i`? 'Constrained 10 of 200 degrees of freedom of u.'

Suggestion: `j - i - mycons[i:j].sum()` is faster and uses less memory at the expense of one non-whitespace character.

examples/cahnhilliard.py
```diff
@@ -114,7 +114,7 @@ def main(size: Length = parse('10cm'),
 
     For this reason, the `stab` enum in this script defines the stabilizing term
     `δψ`, rather than `f`, allowing Nutils to construct the residuals through
-    automatic differentiation using the `optimize` method.
+    automatic differentiation.
```
joostvanzwieten, line 118: Unrelated to the PR, but we're using symbolic differentiation, not automatic differentation.

joostvanzwieten (changes requested):

nutils/solver.py
gertjanvanzwieten: fixed

examples/cahnhilliard.py
gertjanvanzwieten: fixed

gertjanvanzwieten:
New since last push: commits 8bce3811, 8d3334fa, 7c0913e8. Fixed e0cfc4b9 to account for a constant matrix with non-constant residual.

joostvanzwieten (approved):

PR #899, opened 2025-02-24
Fix immutable-making bug in evaluable.compile

gertjanvanzwieten:
This patch limits the generation of v.setflags(write=False) immutability safeguards by the evaluable compiler to variables that originate from Array objects. Formerly all variables were modified, including e.g. Tuple, resulting in an error as arbitrary objects do not offer a setflags method.


gertjanvanzwieten:
Replaced by #900

PR #910, opened 2025-03-27
Explicit spaces function.grad, Namespace.define_for

joostvanzwieten:
The `function.grad` family computes gradients of the geometry to the root coordinate systems of all spaces where the geometry is bound to. Under certain circumstance, e.g. space-time tensor domain with a geometry that rotates in time, the gradient w.r.t. a geometry must be computed w.r.t. the space root coordinates only, not in space and time (which fails to lower anyhow, because the dimension of space and time is larger than the dimension of the geometry). This PR adds support for specifying the spaces in which the gradient must be computed to the `function.grad` family and `Namespace.define_for`.

### DIFF
--- a/tests/test_expression_v2.py
+++ b/tests/test_expression_v2.py
@@ -542,6 +542,18 @@ class Namespace(TestCase):
         with self.assertRaisesRegex(ValueError, 'The curl can only be defined for a geometry with shape'):
             ns.define_for('x', curl='curl')
 
+    def test_define_for_2d_spaces(self):
+        ns = expression_v2.Namespace()
+        X, ns.x0 = mesh.rectilinear([numpy.linspace(0, 1, 3), numpy.linspace(0, 1, 5)], space='X')
+        T, ns.t = mesh.line(1, space='T')
+        ns.δ = function.eye(2)
+        ns.x_i = 'x0_i + δ_i0 t'
+        ns.define_for('x', gradient='∇', normal='n', jacobians=['dV', 'dS'], spaces=iter(X.spaces))
+        self.assertAlmostEqual((X*T).integral('dV' @ ns, degree=2).eval(), 1)
+        self.assertAlmostEqual((X*T).integral('∇_i(x_i^2) dV' @ ns, degree=2).eval(), 3)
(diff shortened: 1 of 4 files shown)


tests/test_function.py
```diff
@@ -897,6 +906,11 @@ def test_grad_3d(self):
         x = function.unravel(function.unravel(x, 0, (2, 2)), 0, (2, 1))
         self.assertEvalAlmostEqual(domain, self.grad(x, x), numpy.eye(4, 4).reshape(2, 1, 2, 2, 1, 2))
 
+    def test_grad_spaces(self):
+        X, x = mesh.line(1, space='X')
+        Y, y = mesh.line(1, space='Y')
+        self.assertEvalAlmostEqual(X * Y, self.grad(x**2 + y, x + y, spaces=iter(X.spaces)), 2 * x)
```
gertjanvanzwieten, line 912: Is there a reason we don't just write spaces=X.spaces? (applies to all tests below as well)

tests/test_expression_v2.py
```diff
@@ -542,6 +542,18 @@ def test_define_for_1d(self):
         with self.assertRaisesRegex(ValueError, 'The curl can only be defined for a geometry with shape'):
             ns.define_for('x', curl='curl')
 
+    def test_define_for_2d_spaces(self):
+        ns = expression_v2.Namespace()
+        X, ns.x0 = mesh.rectilinear([numpy.linspace(0, 1, 3), numpy.linspace(0, 1, 5)], space='X')
+        T, ns.t = mesh.line(1, space='T')
+        ns.δ = function.eye(2)
+        ns.x_i = 'x0_i + δ_i0 t'
+        ns.define_for('x', gradient='∇', normal='n', jacobians=['dV', 'dS'], spaces=iter(X.spaces))
```
gertjanvanzwieten, line 551: Same question here and below

gertjanvanzwieten (approved):

tests/test_function.py
joostvanzwieten: Yes. `function.grad()` and related functions annotate the spaces argument as an iterable. I want to make sure that this is actually true. For instance, `function.laplace()` has to pass `frozenset(spaces)` to `function.grad()` and `function.div()` or the unittest will fail.

PR #911, opened 2025-03-27
Fix hash of  compiled functions

joostvanzwieten:
The hash of compiled functions is based solely on the source code. Constants are currently added to globals of the compiled function under `c{index}`, hence constants are not included in the hash. Furthermore, parameters of evaluables using `evalf` via the `e{index}` instance, also added to the globals, are also not represented in the source code. As a consequence, two compiled functions with different constants or parameters of certain evaluables may have the same hash. This PR fixes the problem by storing constants under `c{hash(constant)}` and disallowing using `e{index}.evalf`.


nutils/evaluable.py
```diff
@@ -4116,6 +4116,9 @@ def _derivative(self, var, seen):
     def evalf(f, sh1, sh2):
         return f.reshape(f.shape[:-1] + (sh1, sh2))
 
+    def _compile_expression(self, py_self, add_constant, f, sh1, sh2):
+        return _pyast.Variable('evaluable').get_attr('Unravel').get_attr('evalf').call(f, sh1, sh2)
```
gertjanvanzwieten, line 4111: The commit message said Ravel, should be Unravel.

nutils/function.py
```diff
@@ -699,10 +697,29 @@ def lower(self, args: LowerArgs) -> evaluable.Array:
...
+            evalargs,
+            self.shape[self._npointwise:],
+            self.dtype,
+            self.spaces,
+            types.frozendict(self._arguments),
+            LowerArgs(points_shape, types.frozendict(args.transform_chains), types.frozendict(coordinates)),
+        )
+
+    @types.hashable_function('NotImplemented')
+    @staticmethod
```
gertjanvanzwieten, line 721: staticmethod is implied by hashable_function

nutils/function.py
```diff
@@ -736,8 +753,9 @@ def evalf(cls, *args: Any) -> numpy.ndarray:
 
         raise NotImplementedError  # pragma: nocover
 
-    @classmethod
-    def partial_derivative(cls, iarg: int, *args: Any) -> IntoArray:
+    @types.hashable_function('NotImplemented')
+    @staticmethod
```
gertjanvanzwieten, line 757: staticmethod is implied by hashable_function

nutils/function.py
```diff
@@ -761,7 +779,7 @@ def partial_derivative(cls, iarg: int, *args: Any) -> IntoArray:
             The partial derivative of this function to the given argument.
         '''
 
-        raise NotImplementedError('The partial derivative of {} to argument {} (counting from 0) is not defined.'.format(cls.__name__, iarg))  # pragma: nocover
+        raise NotImplementedError('The partial derivative to argument {} (counting from 0) is not defined.'.format(cls.__name__, iarg))  # pragma: nocover
```
gertjanvanzwieten, line 782: I see only one set of {} and two format arguments. Also, let's switch to a format string?

nutils/evaluable.py
```diff
@@ -7006,6 +7006,10 @@ def array_iadd(self, acc: _pyast.Expression, inc: _pyast.Expression):
         # Appends statements for incrementing array `acc` with array `inc`.
         self.exec(_pyast.Variable('numpy').get_attr('add').call(acc, inc, out=acc))
 
+    def array_imul(self, acc: _pyast.Expression, inc: _pyast.Expression):
```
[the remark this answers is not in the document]
joostvanzwieten: Applied.

nutils/evaluable.py
```diff
@@ -920,7 +920,7 @@ def _compile(self, builder):
         # convert the array to a `numpy.number` (`self.value[()]`), which
         # behaves like a `numpy.ndarray`, and has much faster implementations
         # for add, multiply etc.
-        return builder.add_constant_for_evaluable(self, self.value[()] if self.ndim == 0 else self.value)
+        return builder.add_constant(self.value[()] if self.ndim == 0 else self.value)
```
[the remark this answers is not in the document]
joostvanzwieten: Good idea. We should return something array-like, though, so this should compile to `numpy.{dtype}_({constant})`. Applied.

nutils/evaluable.py
```diff
@@ -6835,13 +6835,10 @@ def new_empty_array_for_evaluable(self, array: Array) -> _pyast.Variable:
...
-        # Assigns `value` to constant `c{id}` where `id` is the unique index of
-        # `evaluable`. It is an error to assign `value` to `evaluable` multiple
-        # times unless the values are identical.
-        const = _pyast.Variable('c{}'.format(self._get_evaluable_index(evaluable)))
-        if self._globals.setdefault(const.py_expr, value) is not value:
-            raise ValueError('Cannot assign different values to the same constant.')
+    def add_constant(self, value) -> _pyast.Variable:
+        # Assigns `value` to constant `cx{nutils_hash(value)}`.
+        const = _pyast.Variable('c{}'.format(types.nutils_hash(value).hex()))
+        assert self._globals.setdefault(const.py_expr, value) is value
```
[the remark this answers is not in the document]
joostvanzwieten: I think they are, but I'll change this to `==`.

nutils/evaluable.py
```diff
@@ -206,12 +206,12 @@ def _compile(self, builder: '_BlockTreeBuilder') -> _pyast.Expression:
         # calling `Evaluable._compile`, because the former caches the compiled
         # evaluables.
         args = builder.compile(self.dependencies)
-        expression = self._compile_expression(builder.get_evaluable_expr(self), *args)
+        expression = self._compile_expression(builder.get_evaluable_expr(self), builder.add_constant, *args)
```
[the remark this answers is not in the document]
joostvanzwieten: 1. `add_constant` is only used a couple of times. `SearchSorted`, `Monomial` can be rewritten using `Constant`. `Transform*` not, because the constants are not arrays. I'm fine with adding a `TransformsSequenceConstant` or even a general purpose `EvaluableConstant` and dropping the `add_constant` argument from `_compile_expression`.
2. Implementing `_compile` should be done with more care than `_compile_expression`. The order in which dependencies are compiled and blocks are created matters. Proving a `_BlockBuilder` instead of a `_BlockTreeBuilder` would be less of a problem.
3. `_compile_expression` is a convenience function that takes *compiled* arguments. `builder.assign_for` takes care of only half of the convenience. Most evaluables implement `_compile_expression`.

nutils/evaluable.py
joostvanzwieten: Fixed.

nutils/function.py
joostvanzwieten: fixed

nutils/function.py
joostvanzwieten: fixed

nutils/evaluable.py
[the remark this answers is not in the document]
gertjanvanzwieten: The array-like criterion muddies the generated code quite a bit, and adds overhead for often reused scalars. So is this actually a good idea or were we better off before?

PR #914, opened 2025-04-03
Nested integrals

joostvanzwieten:


nutils/function.py
```diff
@@ -18,84 +18,138 @@
...
 
     points_shape: Tuple[evaluable.Array, ...]
-    transform_chains: Mapping[str, Tuple[Transforms, evaluable.Array]]
-    coordinates: Mapping[str, evaluable.Array]
-
-    def consistent(self):
-        return not any(evaluable._any_certainly_different(coords.shape[:-1], self.points_shape)
-            and space in self.transform_chains
-                for space, coords in self.coordinates.items())
+    args: LowerArg
```
gertjanvanzwieten, line 87: Tuple[LowerArg]

nutils/function.py
```diff
@@ -18,84 +18,138 @@
...
     ----------
     points_shape : :class:`tuple` of scalar, integer :class:`nutils.evaluable.Array`
         The shape of the leading points axes that are to be added to the
         lowered :class:`nutils.evaluable.Array`.
-    transform_chains : mapping of :class:`str` to tuples of :class:`nutils.transformseq.Transforms` and :class:`nutils.evaluable.Array`
-        A mapping of spaces to transforms sequences and evaluable indices.
-    coordinates : mapping of :class:`str` to :class:`nutils.evaluable.Array` objects
-        The coordinates at which the function will be evaluated.
+    args : :class:`tuple` or :class:`LowerArg`
+        Tuple of transforms, index and coordinates per space.
```
gertjanvanzwieten, line 83: tuple _of_ LowerArg

nutils/function.py
```diff
@@ -147,9 +140,16 @@ def __getitem__(self, space: str) -> LowerArg:
...
 
-        return tuple(arg.space for arg in self.args)
+        The spaces are ordered by :attr:`args`. Spaces that have another entry
+        in :attr:`args` to the right are omitted.
+        '''
+
+        spaces = []
+        spaces.extend(arg.space for arg in reversed(self.args) if arg.space not in spaces)
+        spaces.reverse()
+        return tuple(spaces)
```
gertjanvanzwieten, line 172: In the situations where the order is used, the next step seems to always be take take the argument from args via args[space], so maybe we should offer something like .exposed_args instead for these situations. Then for spaces we can simply return `{arg.space for arg in self.args}`.

nutils/sample.py
```diff
@@ -1060,7 +1060,7 @@ def __init__(self, integrand: function.Array, sample: Sample) -> None:
         super().__init__(shape=integrand.shape, dtype=float if integrand.dtype in (bool, int) else integrand.dtype, spaces=integrand.spaces - frozenset(sample.spaces), arguments=integrand.arguments)
 
     def lower(self, args: function.LowerArgs) -> evaluable.Array:
-        ielem = evaluable.loop_index('_sample_' + '_'.join(self._sample.spaces), self._sample.nelems)
+        ielem = evaluable.loop_index(f'_sample_{len(args.args)}', self._sample.nelems)
```
gertjanvanzwieten, line 1063: Can you explain this change?

nutils/function.py
```diff
@@ -65,6 +65,14 @@ def map_coordinates(self, map):
 
         return self if self.coordinates is None else self.replace(coordinates=map(self.coordinates))
 
+    def rename_spaces(self, map: Mapping[str, str]) -> 'LowerArg':
+        '''Return a copy of the :class:`LowerArg` with :attr:`space` renamed using ``map``
+
+        If the old space is not found in ``map``, the :class:`LowerArg` is returned unchanged.
+        '''
+
+        return self if (space := map.get(self.space, self.space)) == self.space else self.replace(space=space)
```
gertjanvanzwieten, line 70: I would put the space= assignment on the line above, since there is no reason not to.

nutils/function.py
```diff
@@ -1127,11 +1126,10 @@ def __init__(self, arg: Array, space: str) -> None:
         super().__init__(arg.shape, arg.dtype, arg.spaces, arg.arguments)
 
     def lower(self, args: LowerArgs) -> evaluable.Array:
-        opposite_args = LowerArgs(
-            args.points_shape,
-            tuple(LowerArg(arg.space, arg.transforms_list[::-1], arg.index, arg.coordinates) if arg.space == self._space else arg for arg in args.args),
-        )
-        return self._arg.lower(opposite_args)
+        opp_space = '~' + self._space
+        if self._space in args.spaces and opp_space in args.spaces:
```
gertjanvanzwieten, line 1130: Shouldn't this be 'or' ?

nutils/function.py
joostvanzwieten: fixed

nutils/function.py
joostvanzwieten: fixed

nutils/function.py
joostvanzwieten: applied

nutils/function.py
```diff
@@ -76,6 +80,14 @@ def __or__(self, other: 'LowerArgs') -> 'LowerArgs':
         points_shape = self.points_shape + other.points_shape
         return LowerArgs(points_shape, transform_chains, coordinates)
 
+    def __add__(self, other: 'LowerArgs') -> 'LowerArgs':
+        duplicates = set(self.transform_chains) & set(other.transform_chains)
+        if duplicates:
+            raise ValueError(f'Nested integrals or samples in the same space: {", ".join(sorted(duplicates))}.')
+        if not evaluable._all_certainly_equal(self.points_shape, other.points_shape):
+            raise ValueError('Cannot add `LowerArgs` with points shapes that are not certainly equal.')
```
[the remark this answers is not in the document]
joostvanzwieten: Applied

nutils/function.py
joostvanzwieten: `function.opposite(arg)` wraps `arg` in `_Opposite(arg, space)` for each space in `arg`, regardless if this space has an opposite. This was already the case before this PR. If we'd swap `space` for `~space` if `~space` doesn't exist, `arg` becomes unevaluable. So we need `~space` to exist. Requiring `space` is not necessary, but prevents `opposite(opposite(arg))` to work.

nutils/sample.py
joostvanzwieten: Without this change nested integrals over the same space would get the same loop index.

gertjanvanzwieten: Ah right, and now the index name is the nesting depth and therefore (locally) unique. Got it.

nutils/function.py
gertjanvanzwieten: clear

nutils/function.py
joostvanzwieten: I've add `LowerArgs.exposed` and changed `LowerArgs.spaces` to a `frozenset`, since we're now only using this attribute to test existence of certain spaces (in `_Opposite`).

PR #935, opened 2026-03-19
Units

gertjanvanzwieten:


.github/workflows/test.yaml
```diff
@@ -186,11 +187,12 @@ jobs:
       - name: Install Nutils and dependencies
         id: install
         run: |
-          python -um pip install --upgrade --upgrade-strategy eager wheel
+          python -um pip install --upgrade --upgrade-strategy eager pip wheel
```
joostvanzwieten: I'm not sure if we still need wheel.

.github/workflows/test.yaml
```diff
@@ -186,11 +187,12 @@ jobs:
...
           # Install Nutils from `dist` dir created in job `build-python-package`.
-          python -um pip install "$_wheel[import-gmsh,matrix-scipy,export-mpl]"
+          python -um pip install "$_wheel"
+          python -um pip install --requirements-from-script examples/${{ matrix.example }}.py
```
joostvanzwieten: I think we should merge the two pip installs. Say we're on a release branch and we accidentally backport an increased version requirement on Nutils for some example, then, with two separate installs, the second install happily upgrades Nutils to a new version, if available on PyPI. This is because pip automatically upgrades packages to be installed, excluding their dependencies.

With merged pip installs, however, we get the following error:

```
ERROR: Cannot install nutils 8.9 (from $wheel) and nutils>=9 because these package versions have conflicting dependencies.
```

joostvanzwieten (changes requested):

.github/workflows/test.yaml
gertjanvanzwieten: Indeed, removed

.github/workflows/test.yaml
gertjanvanzwieten: Fixed

joostvanzwieten (approved):

PR #937, opened 2026-08-26
Ruff check

gertjanvanzwieten:


gertjanvanzwieten:
See #938 for failing MKL tests.

joostvanzwieten (dismissed):

src/nutils/evaluable.py
```diff
@@ -4710,64 +4710,62 @@ class Choose(Array):
...
+            return self.choices[0]
+        index, *choices, where = unalign(self.index, *self.choices)
         if len(where) < self.ndim:
-            return align(Choose(index, choices), where, self.shape)
+            return align(Choose(index, tuple(choices)), where, self.shape)
 
     def _multiply(self, other):
         if isinstance(other, Choose) and self.index == other.index:
-            return Choose(self.index, self.choices * other.choices)
+            return Choose(self.index, tuple(c1 * c2 for c1, c2 in zip(self.choices, other.choices)))
```
joostvanzwieten, line 4747: Shouldn't we check here that `self` and `other` have the same number of items?

gertjanvanzwieten: Good catch. Fixed and I also added the equivalent ._add simplification plus unit tests.

gertjanvanzwieten:
Rebased on top of #939 and limited to ruff fixes, remaining changes to follow in a separate PR.

joostvanzwieten (approved):

